Issue #30, opened 2022-05-04
Determine registry to deploy docker images to


koegel:

jfaltermeier:
I'll open a dedicated docker hub account for now. 
After #31 we may then switch to the foundations dockerhub: https://hub.docker.com/u/eclipse/

jfaltermeier:
https://hub.docker.com/u/theiacloud

PR #37, opened 2022-05-13
Determine registry to deploy docker images to #30

jfaltermeier:
* switch to docker hub


lucas-koehler (commented):
Hey, looks pretty good to me already, thanks.
I noticed the Building-Internal document says in the beginning:
>This page contains the build instructions including tagging and pushing the images to the internal test cluster. There are no secrets in this document.

I think this is no longer accurate and should be adapted to say that we push to dockerhub

jfaltermeier:
> Hey, looks pretty good to me already, thanks. I noticed the Building-Internal document says in the beginning:
> 
> > This page contains the build instructions including tagging and pushing the images to the internal test cluster. There are no secrets in this document.
> 
> I think this is no longer accurate and should be adapted to say that we push to dockerhub

Thanks, I've updated the doc

lucas-koehler (approved):

PR #46, opened 2022-06-30
Adjust code to defined terminology

jfaltermeier:
Rename Template to AppDefinition
Rename Workspace to Session
move from alpha to beta

Rename workspace service to Theia.cloud service
Rename operator to Theia.cloud operator

Fixes #20

PR #50, opened 2022-07-05
Allow to change templates without having to compile the operator

### LINKED ISSUE #49: Allow to change the kubernetes templates without having to compile the operator
We have some kubernetes template files that are used by our operator to create all kubernetes resources. 
If a user wants to make subtle changes to these templates we should have a defined way to simply replace those templates without having to rebuild the whole application. 

Proposed solution is to check a dedicated directory for templates. If a template can be found, we will use this one. Otherwise the operator will use its default template. 
This way users may simply add additional files to a dockerimage without having to know Java.

jfaltermeier:
* Update JavaResourceUtil to read templates from directory
* use template from directory if present
* otherwise use default template

fixes #49 

depends on
* https://github.com/eclipsesource/theia-cloud/pull/46


jfaltermeier:
Merged via https://github.com/eclipsesource/theia-cloud/pull/42

PR #61, opened 2022-07-25
Update Try Now page UI

AlexandraBuzila:
* add background
* add new spinner
* update styling for button
* update error message on maximum instances

error_message


node/try-now-page/src/App.tsx
```diff
@@ -37,58 +37,99 @@ function App(): JSX.Element {
...
   };
 
-  const openTerms = (): boolean => {
-    window.open('/terms.html', 'popup', 'width=600,height=600');
-    return false;
-  };
-
   return (
     <div className='App'>
-      <img src={'/logo.png'} className='App__logo' alt='logo' />
```
jfaltermeier, line 47: The logo.png can be influenced via the helm values.

E.g. https://github.com/eclipsesource/theia-cloud/blob/d2cffe63f23db8bcab37025361d5598041b1606d/helm/theia.cloud/valuesMinikubeCDT.yaml#L4

This value is then used here: https://github.com/eclipsesource/theia-cloud/blob/d2cffe63f23db8bcab37025361d5598041b1606d/helm/theia.cloud/templates/landing-page-config-map.yaml#L18

jfaltermeier (approved):
Thank you for the contribution. Looks great and works as expected. 

There is one small regression however. It was possible to exchange the logo.png beforehand. (See inline comments). Since we have switched to a svg this does not work anymore. We should fix this at some point, but for now I will merge this PR. I will open a ticket for the regression.

PR #66, opened 2022-08-12
Add OpenAPI specs for Theia.cloud service

jfaltermeier:
* enable OpenAPI generation
* add annotations
* commit resulting json under doc/docs/openapi.json

Easy way to test would be launching a swagger-ui via docker:
```
docker pull swaggerapi/swagger-ui
docker run -p 80:8080 swaggerapi/swagger-ui
```
then browsing to http://localhost/ and exploring the spec from here: https://raw.githubusercontent.com/jfaltermeier/theia-cloud/jf/restapi/doc/docs/openapi.json

### DIFF
--- a/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/workspace/WorkspaceCreationRequest.java
+++ b/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/workspace/WorkspaceCreationRequest.java
@@ -29,7 +29,7 @@ public class WorkspaceCreationRequest extends ServiceRequest {
     @Schema(title = "The app this workspace will be used with.", required = false)
     public String appDefinition;
 
-    @Schema(title = "The label of the workspace", required = true)
+    @Schema(title = "The label of the workspace", required = false)
     public String label;
 
     public WorkspaceCreationRequest() {
--- a/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/workspace/WorkspaceCreationRequest.java
+++ b/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/workspace/WorkspaceCreationRequest.java
@@ -16,13 +16,20 @@
  ********************************************************************************/
 package org.eclipse.theia.cloud.service.workspace;
 
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.eclipse.theia.cloud.service.ServiceRequest;
 
+@Schema(name = "Workspace Creation Request", description = "Request to create a new workspace.")
 public class WorkspaceCreationRequest extends ServiceRequest {
     public static final String KIND = "workspaceCreationRequest";
 
+    @Schema(title = "The user identification, usually the email address.", required = true)
     public String user;
+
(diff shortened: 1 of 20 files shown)


java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/workspace/WorkspaceCreationRequest.java
```diff
@@ -16,13 +16,20 @@
...
 public class WorkspaceCreationRequest extends ServiceRequest {
     public static final String KIND = "workspaceCreationRequest";
 
+    @Schema(title = "The user identification, usually the email address.", required = true)
     public String user;
+
+    @Schema(title = "The app this workspace will be used with.", required = false)
     public String appDefinition;
+
+    @Schema(title = "The label of the workspace", required = true)
```
martin-fleck-at, line 32: I think the label is actually optional as it can be generated based on the user and the app definition. At least the workspace creation still seems to work without it.

java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/workspace/WorkspaceCreationRequest.java
```diff
@@ -16,13 +16,20 @@
...
 import org.eclipse.theia.cloud.service.ServiceRequest;
 
+@Schema(name = "Workspace Creation Request", description = "Request to create a new workspace.")
 public class WorkspaceCreationRequest extends ServiceRequest {
     public static final String KIND = "workspaceCreationRequest";
 
+    @Schema(title = "The user identification, usually the email address.", required = true)
     public String user;
+
+    @Schema(title = "The app this workspace will be used with.", required = false)
```
martin-fleck-at, line 29: The appDefinition is not strictly needed but I saw that the label and name generation still depend on it, so we now get `null` in the name and label of the workspace which is not a problem, just doesn't look very nice. So maybe we should mark it required for now or ensure that we consider null values in the name and label generation, what do you think?

martin-fleck-at (changes requested):
Hi @jfaltermeier, Thank you very much for that change, I think that will help people to explore the REST API a lot! I only have one minor comment regarding an optional field.

java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/workspace/WorkspaceCreationRequest.java
jfaltermeier: I've changed to label generation in a follow up commit

jfaltermeier:
Thanks for the review. I've added a further commit addressing the comments.

java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/workspace/WorkspaceCreationRequest.java
martin-fleck-at: Perfect, thank you!

martin-fleck-at (approved):
Great, everything looks good to me now, thank you Johannes!

PR #68, opened 2022-08-18
Removed . from chart name

cjsteele:
Periods in the chart name unfortunately cause issues when using theia-cloud as a subchart, and are also against the [helm chart naming conventions ](https://helm.sh/docs/chart_best_practices/conventions/). Dashes can also cause issues (they are not allowed in values, there is a workaround but it's easier to just avoid them) so I would suggest just concatenation of the two words as a name.

I have to keep this changed in the downstream so it would be very convenient for me to have this changed in the main repository to avoid conflicts.

### DIFF
--- a/helm/theia.cloud/Chart.yaml
+++ b/helm/theia.cloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
-name: theia.cloud
-description: A Helm chart for Theia.Cloud
+name: theiacloud
+description: A Helm chart for Theia Cloud
 
 # A chart can be either an 'application' or a 'library' chart.
 #


jfaltermeier:
Thank you for the contribution. 
Could you please sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) ([How to sign](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Contributor_Agreement)) for your email address? This is mandatory before we can integrate your fix. Thanks!

cjsteele:
> Thank you for the contribution. Could you please sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) ([How to sign](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Contributor_Agreement)) for your email address? This is mandatory before we can integrate your fix. Thanks!

I've done the registration and signed the contributor's agreement.

PR #72, opened 2022-08-19
Fix generation of names for k8s resources

### LINKED ISSUE #71: Fix generation of names for k8s resources
Names have to follow these conventions:
* contain at most 63 characters
* contain only lowercase alphanumeric characters or '-'
* start with an alphanumeric character
* end with an alphanumeric character

Also they have to be unique. 

At the moment there may be cases where we have duplicate names, e.g. when generating the names for configmaps:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: ws-1660897383007theia-cloud-demo-john-doe-greatcompany-session
  labels:
    theiacloud: emails
...
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: ws-1660897383007theia-cloud-demo-john-doe-greatcompany-session
  labels:
    theiacloud: proxy
...
```
For these cases the suffix is lost at the moment.

Affected versions are 0.8.0.MS3 and 0.8.0.MS4

jfaltermeier:
* name pattern will be (InstanceId-)CustomResourceUID-Suffix-AdditionalInformation
* If string gets to long we cut from the end. This should only affect additional information which is only used to make the resource purpose easier to grasp on first sight.

Fixes #71

### DIFF
--- a/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
+++ b/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
@@ -20,27 +20,131 @@ import org.eclipse.theia.cloud.common.k8s.resource.Session;
 import org.eclipse.theia.cloud.common.k8s.resource.Workspace;
 
 public final class NamingUtil {
+
     public static final int VALID_NAME_LIMIT = 62;
+
+    private static final int MAX_IDENTIFIER_LENGTH = 24;
 
     private NamingUtil() {
 	// utility
     }
 
-    public static String createName(AppDefinition appDefinition, int instance, String suffix) {
-	return createName(instance + "-" + appDefinition.getMetadata().getUid(), suffix,
-		getAdditionalInformation(appDefinition));
+    /**
+     * Creates a string that can be used as a name for a kubernetes object.
+     * 
+     * When the same arguments are passed to this method again, the resulting name
+     * will be the same.
+     * 
+     * For different arguments the resulting name will be unique in the cluster.
+     * 
+     * @param appDefinition the {@link AppDefinition}
+     * @param instance      instance id
+     * @param identifier    a short description/name of the kubernetes object for
+     *                      which this name will be used. This will be part of the
+     *                      unique sections of the generated name. <b>The combined
+     *                      length of the instance and the identifier must be at
+     *                      most 23 characters long!</b>
+     * @return the name
+     */
+    public static String createName(AppDefinition appDefinition, int instance, String identifier) {
+	/*
+	 * Kubenertes UIDs are standardized UUIDs/GUIDs. This means the uid string will
+	 * have a length of 36.
+	 * 
+	 * Unique part of the name will consist of the instance followed by the app
+	 * definition's uuid followed by the identifier. Parts will be separated with
+	 * "-". This must be shorter than {@link NamingUtil.VALID_NAME_LIMIT}
+	 * 
+	 * We fill remaining space with additional information about the app definition.
+	 * This may be trimmed away however.
+	 */
+	String prefix = instance + "-";
+	return createName(prefix + appDefinition.getMetadata().getUid(), identifier,
+		getAdditionalInformation(appDefinition), MAX_IDENTIFIER_LENGTH - prefix.length());
     }
 
-    public static String createName(Session session, String suffix) {
-	return createName(session.getMetadata().getUid(), suffix, getAdditionalInformation(session));
+    /**
+     * Creates a string that can be used as a name for a kubernetes object.
+     * 
+     * When the same arguments are passed to this method again, the resulting name
+     * will be the same.
+     * 
+     * For different arguments the resulting name will be unique in the cluster.
+     * 
+     * @param session    the {@link Session}
+     * @param identifier a short description/name of the kubernetes object for which
+     *                   this name will be used. This will be part of the unique
+     *                   sections of the generated name. <b>Must be at most 24
+     *                   characters long!</b>
+     * @return the name
+     */
+    public static String createName(Session session, String identifier) {
+	/*
+	 * Kubenertes UIDs are standardized UUIDs/GUIDs. This means the uid string will
+	 * have a length of 36.
+	 * 
+	 * Unique part of the name will consist of the sessions's uuid followed by the
+	 * identifier. Parts will be separated with "-". This must be shorter than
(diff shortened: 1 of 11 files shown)


java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
```diff
@@ -15,13 +15,49 @@
...
+
+    public static String createName(Session session, String suffix) {
+	return createName(session.getMetadata().getUid(), suffix, getAdditionalInformation(session));
+    }
+
+    public static String createName(Workspace workspace, String suffix) {
+	return createName(workspace.getMetadata().getUid(), suffix, getAdditionalInformation(workspace));
+    }
+
+    private static String createName(String prefix, String suffix, String additionalInformation) {
```
martin-fleck-at, line 42: Nit: Could probably made generic and public with "-" being made into a generic delimiter constant:
```
private static String createName(String... parts) {
   return asValidName(String.join("-", parts));
 }
 ```

File: java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java [hunk 2]
```diff
@@ -15,13 +15,49 @@
...
+
 public final class NamingUtil {
     public static final int VALID_NAME_LIMIT = 62;
 
     private NamingUtil() {
 	// utility
     }
 
+    public static String createName(AppDefinition appDefinition, int instance, String suffix) {
+	return createName(instance + "-" + appDefinition.getMetadata().getUid(), suffix,
```
martin-fleck-at, line 30: Nit: This will only work because we implicitly know that the Java-generated Uid is 36 characters long leaving us 27 (63-36) characters for the instance and the "suffix". Since we control all the suffixes at the moment we know that none of them are currently too long to go over the 63 character limit as otherwise we would not generate unique enough names. I think this should also be commented on the method to ensure that we do not forget this implicit restriction.

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
```diff
@@ -15,13 +15,49 @@
...
+		getAdditionalInformation(appDefinition));
```
martin-fleck-at, line 31: Nit: The additional info is called additional info because we do not need it to make up a unique name but we want to add as much info in the name as possible, is that right? If so, I'd suggest to comment that somewhere (Javadoc or in this method) to ensure we do not forget that they are not strictly part of the unique name.

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
jfaltermeier: I've updated the method to use String.join and also validate the length of the non-uuid part for the name. 
However I think it should stay private, because otherwise we can't be sure that the results will actually be unique.

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
```diff
@@ -15,13 +15,49 @@
...
+import org.eclipse.theia.cloud.common.k8s.resource.Workspace;
+
 public final class NamingUtil {
     public static final int VALID_NAME_LIMIT = 62;
 
     private NamingUtil() {
 	// utility
     }
 
+    public static String createName(AppDefinition appDefinition, int instance, String suffix) {
```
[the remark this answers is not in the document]
jfaltermeier: I picked identifier, since this part of the name can be used to identify which kind of kubernetes object this is.

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
jfaltermeier: I've added javadoc/comments.

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
jfaltermeier: I've added a comment about this

jfaltermeier:
Thank you for the review, I've updated the PR.

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
martin-fleck-at: Sure, that is fine by me!

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
martin-fleck-at: Thank you!

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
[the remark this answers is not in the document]
martin-fleck-at: I like it, good name, thanks!

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
martin-fleck-at: Perfect, thanks!

martin-fleck-at (approved):
Thank you for the swift update, all looks good to me now!

jfaltermeier:
Sorry, I squeezend in one more commit which just adds some additional logging, so that we get the final yaml files that are applied. Could you please may check this one as well?

martin-fleck-at (approved):
No problem, code still looks good to me. I did not execute the new code but it does look very safe.

PR #79, opened 2022-08-26
Add ping to service and check from try now page and fix REST API

martin-fleck-at:
Add ping to service and check from try now page
- Update local try-now page config (missing useEphemeralStorage)
- Introduce ping request
- add launch config to build and install the common maven package
- update OpenAPI and regenerate typescript client
- show maintenance information on try now if service is taken offline

Ensure correct execution of GET requests in REST-API
- Do not use JSON body for GET requests but request parameters
- Change to fetch-based API instead of axios to avoid dependency
- Fix incorrect type of WorkspaceListRequest in common
- Fix NPE in appId evaluation


jfaltermeier (approved):
Thanks, LGTM!

PR #80, opened 2022-08-29
Add request timeout to client API and update package versions

martin-fleck-at:


node/common/src/client.ts
```diff
@@ -107,10 +109,10 @@ export namespace TheiaCloud {
     }
   }
 
-  export async function launch(request: LaunchRequest, retries = 0): Promise<void> {
+  export async function launch(request: LaunchRequest, retries = 0, timeout?: number): Promise<void> {
     const launchRequest = { kind: LaunchRequest.KIND, ...request };
     try {
-      const response = await rootApi(request.serviceUrl).servicePost({ launchRequest }, createConfig());
+      const response = await rootApi(request.serviceUrl).servicePost({ launchRequest }, createConfig(timeout || request.timeout));
```
jfaltermeier, line 115: The `request.timeout` is used by the Kubernetes Client. When the timeout is reached we set an error message on the session. I don't think this should be used here.

jfaltermeier (commented):
Thank you for the PR. Looks good overall. 
I have one small comment however.

node/common/src/client.ts
martin-fleck-at: Fixed.

martin-fleck-at:
@jfaltermeier Thank you for the review, I fixed it in a follow-up commit.

jfaltermeier (approved):
Thanks, LGTM!
The npm-packages are not yet released, so at the moment building the activity tracker sample fails.

PR #81, opened 2022-08-31
Check return value of CountDownLatch#await / Fix launch response

jfaltermeier:


martin-fleck-at (approved):
Change looks good to me, thank you Johannes! It seems to me that this change also supersedes PR https://github.com/eclipsesource/theia-cloud/pull/68 is that right?

jfaltermeier:
At least with the chart name. I hope the . in the description is fine

PR #83, opened 2022-09-07
Improve REST API

martin-fleck-at:
Ensure list session request only reports user-specific sessions

Switch from custom error responses to REST error codes
- Introduce custom TheiaCloud errors
- Introduce custom TheiaCloud exception that is mapped to errors
- Remove custom error responses and regenerate REST client

Fix problem with Theia activity tracker
- Switch back to axios-based client API to avoid dependency issues
-- 'fetch' may not be available on node backend of Theia


java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/BaseResource.java
```diff
@@ -29,8 +32,14 @@ public BaseResource() {
 	logger = Logger.getLogger(getClass().getSuperclass());
     }
 
-    protected boolean isValidRequest(ServiceRequest request) {
-	return request != null && request.appId != null && request.appId.equals(appId);
+    protected String evaluateRequest(ServiceRequest request) {
+	String correlationId = generateCorrelationId();
+	if (request == null || request.appId == null || !request.appId.equals(appId)) {
+	    String methodName = Thread.currentThread().getStackTrace()[2].getMethodName();
```
jfaltermeier, line 38: Isn't the `[2]` too brittle? I think this fails if e.g. two method would delegate to a shared util method, which then evaluates the request.

java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/TheiaCloudWebException.java
```diff
@@ -0,0 +1,77 @@
...
+	return workspace;
+    }
+
+    public static WorkspaceSpec throwIfErroneous(WorkspaceSpec spec) {
+	throwIfError(spec.getError());
+	return spec;
+    }
+
+    public static void throwIfError(String error) {
+	if (error != null) {
```
jfaltermeier, line 69: I think we should also check if the error is blank to be consistent with `SessionSpec.hasError()` and `WorkspaceSpec.hasError()`. 
Maybe we should actually use those `hasError()` methods above?

node/common/src/client.ts
```diff
@@ -99,27 +100,22 @@ export namespace TheiaCloud {
...
-      await rootApi(request.serviceUrl).serviceAppIdGet(fullRequest, createConfig());
+      return getData(() => rootApi(request.serviceUrl).serviceAppIdGet(request.appId, createConfig()));
     } catch (error) {
       console.error((error as any).message);
       throw error;
     }
   }
 
-  export async function launch(request: LaunchRequest, retries = 0, timeout?: number): Promise<void> {
+  export async function launch(request: LaunchRequest, retries = 0, timeout?: number, redirect = true): Promise<void> {
```
jfaltermeier, line 112: Should we return the URL here? If redirect is false we might want to receive the URL

jfaltermeier (commented):
Thank you. I have some small comments after the code review. 
I will start testing now.

node/common/src/client.ts
Hadsch: Yes this would be good ;)

jfaltermeier:
I've verified launching with both our landing pages and checked the activity reporting, which works fine again with the generated code using axios. 
For the rest API I've tested listSessions, which now returns the sessions of the user provided in the path

java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/BaseResource.java
martin-fleck-at: Good point! I removed the methodName and log the request kind instead, together with a trace of the actual request.

java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/TheiaCloudWebException.java
martin-fleck-at: Good catch! I unified the handling for these methods.

node/common/src/client.ts
martin-fleck-at: I totally agree, I created a separate method `launchAndRedirect` that includes the redirection in a successful launch case. I also added the retry mechanism to all our REST client methods for further convenience.

martin-fleck-at:
@jfaltermeier Thank you for the fast review, I pushed an update which should address all concerns.

node/common/src/client.ts
```diff
@@ -99,75 +106,56 @@ export namespace TheiaCloud {
...
-        launch(request, retries - 1, timeout);
-      } else {
-        throw error;
-      }
-    }
+    return call(() => rootApi(request.serviceUrl).servicePost(launchRequest, createConfig(timeout)), retries);
+  }
+
+  export async function launchAndRedirect(request: LaunchRequest, timeout?: number, retries?: number): Promise<string> {
+    const url = await launch(request, retries, timeout);
```
jfaltermeier, line 119: Timeout and retry are switched up here

jfaltermeier (changes requested):

node/landing-page/src/components/SessionLauncher.vue
```diff
@@ -40,37 +40,17 @@ export default defineComponent({
...
-            timeout: 300000
-          }
-        )
-        .then(response => {
-          if (response.data.success) {
-            console.log('Redirect to : https://' + response.data.url);
-            location.replace('https://' + response.data.url);
-          } else {
-            this.text = response.data.error;
+      TheiaCloud.launchAndRedirect(this.useEphemeralStorage
```
jfaltermeier, line 43: We are not passing the user email anymore which results in an anonymous user all the time.

martin-fleck-at: You are right, I pushed another fix, thank you for your thorough review!

jfaltermeier (approved):
Thanks, lgtm!

Issue #84, opened 2022-09-15
Triggering TheiaCloud.launch in node API during timeout leads to empty response


martin-fleck-at:
We should check in the REST service whether we already have a pending session or not. Maybe simply calling launch again if there is no error and no URL yet is enough.

PR #86, opened 2022-09-27
Do not return empty URL if same session is requested during timeout

martin-fleck-at:
Fixes https://github.com/eclipsesource/theia-cloud/issues/84


jfaltermeier (approved):
LGTM!

Issue #87, opened 2022-09-28
Allow templates to be used in values file


martin-fleck-at:
In general, the value files in Helm [do not support the template syntax](https://github.com/helm/helm/issues/2492), therefore the values that adopters can provide are never expanded. This is fine if everything is defined in a single fine where [YAML Anchors](https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors) may also be used but there might be use cases when the template syntax may prove useful.

However, we can get support for this by [using the tpl function](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function) on all user-provided values. Basically, they are no longer interpreted as given but treated as a template first. Basically what we need to change in our templates is `image: {{ .Values.image.name }}` to `image: {{ tpl (.Values.image.name | toString) . }}`.

The following cases need to be considered:
- Value without template is used, e.g., `name: theiacloud/theia-cloud-demo:0.8.0.MS6` > value is taken as is.
- Value with template in the beginning is used, e.g.,  `name: "{{ .Values.global.base }}/theia-cloud-demo:0.8.0.MS6"` or `name: "{{ .Values.global.demoImage }}"` > quoting needs to be used as otherwise we do not get a valid YAML file.
- Value with template in other position is used, e.g., `name: theiacloud/{{ .Values.global.baseName }}-demo:0.8.0.MS6` > no quoting necessary.
- Value with missing template is used > missing template is simply treated as an empty string and the normal Kubernetes validation will issue an error if a required field is missing.

PR #88, opened 2022-10-03
Allow templates to be used within values file by using tpl function

martin-fleck-at:
Fixes https://github.com/eclipsesource/theia-cloud/issues/87


jfaltermeier (approved):
LGTM

PR #93, opened 2022-10-12
Set runtime user and group only for application container

lucas-koehler:
Before, we set the runtime user and group for all containers in workspace pods. This included the OAuth proxy.
This could lead to undesired behavior when running the OAuth proxy or other sidecars with the app's user id.

This commit moves the user and group configuration down to the actual application container.


jfaltermeier (approved):
Thanks, lgtm!

PR #99, opened 2022-10-30
Pin landing-page Docker container to node 16.x

qmfrederik:
Make upgrading from one node LTS version to the next an explicit action by using a specific LTS version in the Dockerfile.

Use node:16-alpine to ensure we get nodejs 16.x, the current source does not compile using node 18.x.

### DIFF
--- a/dockerfiles/landing-page/Dockerfile
+++ b/dockerfiles/landing-page/Dockerfile
@@ -1,5 +1,5 @@
 # build stage
-FROM node:lts-alpine as build-stage
+FROM node:16-alpine as build-stage
 WORKDIR /app
 COPY node/landing-page/package*.json ./
 RUN yarn install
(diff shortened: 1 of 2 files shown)


jfaltermeier:
Thank you for the contribution. 
Could you please sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) ([How to sign](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Contributor_Agreement)) for your email address? This is mandatory before we can integrate your fix. 
This should make the eclipsefdn/eca check above happy. Ad far as I know the check takes the email-address from the commit, if this is different from the email-address of the Github account.

lucas-koehler:
In the meantime, this was already fixed on the `main` branch. Thank you for the contribution anyway :)

PR #103, opened 2022-11-09
Only configure service's keycloak parameters for enabled keycloak

lucas-koehler:
Improve the service configmap template to only contain keycloak parameters (e.g. the server url) if keycloak is enabled. This prevents empty values overriding the docker image's defaults. Before this change, this could lead to  errors in the service.

Additionally, no longer hand in property `quarkus.oidc.enabled` at runtime to the service because this property can only be set at build time.

Remove workaround keycloak values from `valuesGKETryNow.yaml` again.

## Note for testing

You can test the generation of the generated kubernetes resources with the `helm template` command without needing to deploy the chart.
E.g.: `helm template ./helm/theia.cloud --values ./helm/theia.cloud/valuesGKETryNow.yaml > test.yaml`


jfaltermeier (approved):
LGTM. Checked locally with and without keycloak

PR #108, opened 2022-11-21
Same base url for all components

lucas-koehler:
Facilitate running all of theia cloud's components on the same base url on different paths.

* Adjust ingress configurations
* Adjust operator
* Adjust try now page and common api

TODO:
* keycloak ingress
* operator ingress creation via template

### DIFF
--- a/doc/docs/Install.md
+++ b/doc/docs/Install.md
@@ -18,6 +18,22 @@ As of writing this guide the installation command looks like this:\
 Follow the installation guide for your platform:\
 https://kubernetes.github.io/ingress-nginx/deploy/
 
+### Global certificate (Only when using paths)
+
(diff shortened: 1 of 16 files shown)


jfaltermeier:
I will create the docker images for some initial testing

jfaltermeier:
Regarding the TODOs: 
Keycloak ingress is out of scope for Theia Cloud itself. We may offer a guide on how to do this at some point, but it is not needed for now. 
Regarding the Ingress Creation via Operator I created a ticket to remove this: https://github.com/eclipsesource/theia-cloud/issues/117
This should be done as a follow up, since we have some open PRs building on top of this PR

jfaltermeier (approved):
LGTM! 
I've create one additional follow-up: https://github.com/eclipsesource/theia-cloud/issues/118

PR #111, opened 2022-11-23
Add metrics server endpoint

sgraband:


jfaltermeier (approved):
Thanks. Looks good to me!
I've created two follow up tickets that need to be addressed before we may release this:
https://github.com/eclipsesource/theia-cloud/issues/119
https://github.com/eclipsesource/theia-cloud/issues/120
However we may merge this PR already

PR #115, opened 2022-11-29
Update AppDefinitions for CDT Cloud, Coffee Editor, Theia

martin-fleck-at:
- Re-use already configured host instance in app definitions
- Re-use already configured ingress name in app definitions
- Use 500M as requested memory for CDT Cloud as it should be sufficient


jfaltermeier (approved):
Thanks, and sorry for the miss click close ;-)

PR #116, opened 2022-12-01
Make imagePullPolicy configurable via helm chart and app definition

### LINKED ISSUE #109: Make imagePullPolicy configurable via helm chart

lucas-koehler:
### Make application imagePullPolicy configurable in app definitions 

- Extend appdefinition CRD with imagePullPolicy property
- Extend operator to set this for new session pods and provide a fallback value if not specified

### Make containers' imagePullPolicy configurable via helm chart 

- Add root level value `imagePullPolicy` and a value for each component
- Use component-specific value if available, otherwise fall back to root level property
- Extend templates for this
- Extend app definition CRD to store the pull policy
- Add root level property to all values files. Use Always for online and IfNotPresent for Minikube

Fixes #109 

### DIFF
--- a/helm/theia.cloud/crds/appdefinition-spec-resource.yaml
+++ b/helm/theia.cloud/crds/appdefinition-spec-resource.yaml
@@ -25,6 +25,9 @@ spec:
                   type: string
                 image:
                   type: string
+                imagePullPolicy:
+                  type: string
+                  enum: ["Always", "IfNotPresent", "Never"]
                 pullSecret:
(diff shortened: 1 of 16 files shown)


jfaltermeier:
The main problem with this would probably be that if someone has the old `v3beta` crd installed already, the new custom resources don't match this.

jfaltermeier (approved):

jfaltermeier:
We will bump the versions before a release

PR #121, opened 2022-12-06
Fix vue landing page to work with paths

lucas-koehler:
- Configure the base url to be relative. With that the landing page can run on any path
- Improve variable names in common client and remove obsolete code

Note: Like the Try Now page, the landing page only works if the URL ends with a trailing `/`.

Fix #118

Contributed on behalf of STMicroelectronics

Signed-off-by: Lucas Koehler <lkoehler@eclipsesource.com>

PR #123, opened 2022-12-06
Add fair use to try now terms & conditions and format them

lucas-koehler:
- Apply code formatting to the Try Now page's terms and conditions (via VSCode auto format on save)
- Add fair use to the Try Now page's terms and conditions

### DIFF
--- a/node/try-now-page/public/terms.html
+++ b/node/try-now-page/public/terms.html
@@ -38,7 +38,31 @@
     exploring the features of the modeling IDE.
   </p>
 
-  <h2>3. Disclaimer</h2>
+  <h2>3. Fair use</h2>
+  <p>
+    Theia.Cloud may be used for testing, demoing and showcasing deployed Theia-based tools in a legal and fair manner.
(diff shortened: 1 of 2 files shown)


jfaltermeier:
The terms still use "Theia.Cloud" as a name, however I would suggest to keep it this way for now. 
We will then do a full rename to "Theia Cloud" over all components. 

However I noticed one more thing that slipped through earlier. When using the Try Now Page with Paths, the popup leads to a 404. I could imagine this is a similar fix as in https://github.com/eclipsesource/theia-cloud/pull/121

ksnip_20221206-132107

lucas-koehler:
@jfaltermeier  Good find! I only looked at the updated terms.html directly before.
I added a fix to make the terms.html link relative.
> The terms still use "Theia.Cloud" as a name, however I would suggest to keep it this way for now.
We will then do a full rename to "Theia Cloud" over all components.

I agree :)

jfaltermeier (approved):
Thanks, LGTM!

PR #125, opened 2022-12-09
Add node common API for metrics endpoints

lucas-koehler:
Regenerate the OpenAPI specification and based on this the service's typescript API.

### DIFF
--- a/node/common/src/client/common.ts
+++ b/node/common/src/client/common.ts
@@ -142,7 +142,7 @@ export const toPathString = function (url: URL) {
  */
 export const createRequestFunction = function (axiosArgs: RequestArgs, globalAxios: AxiosInstance, BASE_PATH: string, configuration?: Configuration) {
     return <T = unknown, R = AxiosResponse<T>>(axios: AxiosInstance = globalAxios, basePath: string = BASE_PATH) => {
-        const axiosRequestArgs = {...axiosArgs.options, url: (configuration ? configuration.basePath : basePath) + axiosArgs.url};
+        const axiosRequestArgs = {...axiosArgs.options, url: (configuration?.basePath || basePath) + axiosArgs.url};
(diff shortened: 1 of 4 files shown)


node/common/src/client/common.ts
```diff
@@ -142,7 +142,7 @@ export const toPathString = function (url: URL) {
  */
 export const createRequestFunction = function (axiosArgs: RequestArgs, globalAxios: AxiosInstance, BASE_PATH: string, configuration?: Configuration) {
     return <T = unknown, R = AxiosResponse<T>>(axios: AxiosInstance = globalAxios, basePath: string = BASE_PATH) => {
-        const axiosRequestArgs = {...axiosArgs.options, url: (configuration ? configuration.basePath : basePath) + axiosArgs.url};
+        const axiosRequestArgs = {...axiosArgs.options, url: (configuration?.basePath || basePath) + axiosArgs.url};
```
jfaltermeier, line 145: FYI: I know that at some point this auto-generated code led to errors, however I can't remember all the details. 
At the moment I can't see issues during compilation and we've updated Node Versions in the Dockerfiles, so this might be resolved in the meantime.

jfaltermeier (commented):
Under `common/src/client/*` we have the auto-generated part of the code. However users will usually work with '`common/client.ts` which is written manually and delegates to the generated API. I think we should add a convenience method for calling the performance API there as well.

node/common/src/client/common.ts
lucas-koehler: The optional chaining operator `?` was not supported by Node before version 14. I suspect that this was the issue back then.

lucas-koehler:
@jfaltermeier Thanks for the review! I added a convenience method for the performance endpoint

jfaltermeier (approved):

PR #126, opened 2022-12-12
Adapt service role to remove write access on pods

### LINKED ISSUE #120: Adapt service role to not allow pod creation
See https://github.com/eclipsesource/theia-cloud/pull/111 (https://github.com/eclipsesource/theia-cloud/pull/111/files#diff-399a4c60b9a4c2087e9f8d732be5e32c97a7a23706ecbb507990b706343e22ba)

For checking the metrics we need to access the pods. The service should only be able to modify custom resources. 
We have to slit the roles here.

lucas-koehler:
Split the rules for custom resource and metrics access. Now, the service can no longer create, modify or delete pods.

Fixes #120

### DIFF
--- a/helm/theia.cloud/templates/service-role.yaml
+++ b/helm/theia.cloud/templates/service-role.yaml
@@ -8,13 +8,18 @@ rules:
     apiGroups:
       - ""
       - theia.cloud
-      - metrics.k8s.io
     resources:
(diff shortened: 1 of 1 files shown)


jfaltermeier (approved):
Thanks, lgtm! Tested with postman and call is still working

PR #129, opened 2022-12-13
Adapt OAuth proxy config map for path based deployment

### LINKED ISSUE #128: OAuth proxy config map is not adapted for path based deployment
In `oauth2-configmap-oauth2proxy-keycloak.yaml`, the config's properties `cookie_domains` and `whitelist_domains` must contain the base path for path based deployments. So far, they only contain `.Values.hosts.instance` that should only be used for subdomain based deployments. Otherwise, the OAuth proxy of session pods will not authorize the user.

lucas-koehler:
Fix #128


jfaltermeier (approved):
Thanks, I was able to reproduce the issue and confirmed the fix.

Issue #18, opened 2022-04-06
Implement unit tests


koegel:

jfaltermeier:
Keep open

lucas-koehler:
Regarding the service, initial setup and a unit test that shows how to test resources in the service is available on this branch already: [feat/basic-authorization](https://github.com/eclipsesource/theia-cloud/tree/feat/basic-authorization)

askarbozcan:
Keep open please, operator really needs unit tests IMO.

PR #141, opened 2023-02-08
Add basic authorization and unit tests to the service

lucas-koehler:
This PR adds basic authorization for the delete workspace and stop session endpoints. That is, only the creator of a workspace/session is allowed to remove them again. Anonymous users cannot remove any at all.

Additionally, this PR sets up unit tests for the service and adds tests for the new authentication and further security relevant classes.

**Note:** I think this PR should not be squashed to have clearer messages when using git blame later.

Part of #18

# Change details

## Ensure users can only delete their own workspaces and sessions
- Introduce TheiaCloudUser to represent a user independent of auth details
- Produce TheiaCloudUser to make it injectable based on the authenticated user (or none if keycloak is not used)
- Check that sessions and workspaces belong to the requesting user before deleting them. Otherwise return a 403 Forbidden error

## Test setup and refactorings
- Add maven dependencies
- Refactor K8sUtil to be injected instead of a static util class. This allows mocking it for tests
- Rename TheiaCloudUser#getName to getIdentifier
- Extract command line application properties to new class `ApplicationProperties` to allow mocking them for tests

## Add tests
- Add WorkspaceResource#delete unit tests
- Add SessionResource tests
- Add TheiaCloudUserProducer tests
- Add unit tests for authenticating anonymous users when keycloak is disabled

## CI
- Add new job to build the service and run the maven tests
- Only execute the service's docker build CI after the tests succeeded
- Skip unit tests during service docker build

## Misc
- Add testing page based on [CRA](https://create-react-app.dev/) to facilitate manual service tests 
- Add README notes for test naming conventions
- Improve Minikube global certificate documentation

# How to test
- Have at least two users in Keycloak
- Use the new testing page to create a workspace and a session with one user
- Deleting both the session as  well as the workspace should return a `403` error when done with the other user
- Deleting as an anonymous user should return a `401` error
- Deleting with the original creator should work

# Attribution


sgraband (commented):
Works great and looks good. I just have online comment:

lucas-koehler:
@sgraband Thanks for your review. I added a comment and replaced the URL with a subdomain-based one to make it clearer.

sgraband (approved):
LGTM :+1: Thank you!

Issue #92, opened 2022-10-12
Review default workspace persistent volume creation to fit production clusters


lucas-koehler:
# Current Situation
We have two operator cloud provider modes configured in the Helm values via `operator.cloudProvider`.
The acceptable values are `K8S` - which suggests the default - and `GKE` (Google Kubernetes Engine).

These modes differ in creating persistent volume claims for persisted workspaces via subclasses of `PersistentVolumeCreator`.

The default implementation in `DefaultPersistentVolumeCreator` first creates a persistent volume (PV) and then the corresponding persistent volume claim (PVC). Thereby, the PV is created on the cluster node's local filesystem via the `hostPath` property. While this makes sense for Minikube, it is not recommeded to use in production:
> In this exercise, you create a hostPath PersistentVolume. Kubernetes supports hostPath for development and testing on a single-node cluster. A hostPath PersistentVolume uses a file or directory on the Node to emulate network-attached storage.
In a production cluster, you would not use hostPath.

Source: https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolume

Thus, this is not suitable for production.

The `GKEPersistentVolumeCreator` just creates a PVC of storage class `default`. This is supported by Google and Azure and probably also Amazon.

# Suggestion:
The GKE approach should be the default and we should use this in operator mode `K8S`. In addition, we add an operator mode `Minikube` that is used for testing configuration.

In addition, we introduce a new Helm value `operator.storageClass` that allows to configure the storage class of PVCs. This allows cluster admins to define a new StorageClass to their needs that is then used by the operator to create the persistent storage for workspaces.
With this, we enable cluster admins to customize workspace storage with no code modifications and a relatively small implementation effort in Theia.Cloud.

jfaltermeier:
If a cluster admin needs even more control than `operator.storageClass` we may add more helm values once needed and they still have the option to use/provide a custom `PersistentVolumeCreator`-implementation. 
So I think with the proposed approach we have good defaults while allowing customizations.

lucas-koehler:
keep open

lucas-koehler:
Fixed via https://github.com/eclipsesource/theia-cloud/pull/144 and https://github.com/eclipsesource/theia-cloud-helm/pull/12

PR #144, opened 2023-02-24
Storage class and amount configuration for workspace volumes

lucas-koehler:
This changes the default persistent volume creation to use a storage class. Before, this was only done when configuring the GKE cloud provider. Now, the K8S cloud provider behaves this way and an additional cloud provider MINIKUBE was added to use the former behavior of the K8S provider. With this, the GKE cloud provider is obsolete and was deleted.

In addition, the used storage class and the amount of requested storage can now be configured via helm values.

- Add arguments `--storageClassName` and `--requestedStorage` to the operator
- Make persistent storage claiming via storage class the default
- Use the former default - explicitly creating the persistent volume - for new cloud provider MINIKUBE only
- Delete cloud provider GKE
- Rename files to fit the new default and cloud provider
- Adapt values files
- Extract persistent volume template replacements to injectable class and bind it in the theia cloud DI module.
- Create reusable TemplateReplacements interface

Part of #92

---

The Minikube part can be tested locally. Simply checkout the corresponding branch of the helm repository (see PR https://github.com/eclipsesource/theia-cloud-helm/pull/12) and use the `helm/theia.cloud/valuesMinikube.yaml` file from this repository.
In addition:
- set ephemeral storage to false and set `operator.requestedStorage` to some value (e.g. 100Mi).
- Build the operator from this branch and configure it in the values
- Launch a new session and verify that the created persistent volume (pv) and persistent volume claim (pcv) each request the configured storage.

Testing the K8S config might be a bit harder. As it's just a replacement more, maybe it is enough to carefully review that part of the code for now (template and replacements)

### DIFF
--- a/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/impl/DefaultPersistentVolumeCreator.java
+++ b/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/impl/DefaultPersistentVolumeCreator.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2022 EclipseSource and others.
+ * Copyright (C) 2022-2023 EclipseSource, STMicroelectronics and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -27,7 +27,7 @@ import org.apache.logging.log4j.Logger;
 import org.eclipse.theia.cloud.common.k8s.client.TheiaCloudClient;
 import org.eclipse.theia.cloud.common.k8s.resource.Workspace;
 import org.eclipse.theia.cloud.operator.handler.PersistentVolumeCreator;
-import org.eclipse.theia.cloud.operator.handler.util.TheiaCloudPersistentVolumeUtil;
+import org.eclipse.theia.cloud.operator.handler.PersistentVolumeTemplateReplacements;
 import org.eclipse.theia.cloud.operator.util.JavaResourceUtil;
 
 import com.google.inject.Inject;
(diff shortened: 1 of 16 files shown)


java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/impl/DefaultPersistentVolumeCreator.java
```diff
@@ -27,46 +27,44 @@
...
+ * Creates persistent volume claims using a storage class. Kubernetes creates
+ * the persistent volume based on the storage class's configuration on the
+ * cluster. Thus, explicit creation of a persistent volume is not required.
+ */
 public class DefaultPersistentVolumeCreator implements PersistentVolumeCreator {
 
     private static final Logger LOGGER = LogManager.getLogger(DefaultPersistentVolumeCreator.class);
 
-    protected static final String TEMPLATE_PERSISTENTVOLUME_YAML = "/templatePersistentVolume.yaml";
+    protected static final int THEIA_CONTAINER_INDEX = 1;
```
jfaltermeier, line 47: Is this used anywhere?

jfaltermeier, line 47: Can we remove this?
I saw that this is coming from the old `GKEPersistentVolumeCreator` but I think this wasn't used before.

jfaltermeier (commented):
Code looks good, I still need to do some testing. 
I have one question inline.

jfaltermeier:
And btw I did a rebase to resolve a conflict.

jfaltermeier (approved):

PR #153, opened 2023-03-29
trynow: Show login button instead of launch button while logged out

lucas-koehler:
- Refactor terms and conditions into a separate component
- Add login button component
- Simplify launch app component
- Simplify calculation of the need to show login button and information
- Fixed wording of the login info

### DIFF
--- a/node/try-now-page/src/App.tsx
+++ b/node/try-now-page/src/App.tsx
@@ -207,7 +207,7 @@ function App(): JSX.Element {
                     />
                 }
               </p>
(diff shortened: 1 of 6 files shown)


jfaltermeier (approved):
Thanks, looks good to me. I've tested with and without keycloak enabled. 

The only thing I've noticed is that when I accept the the Terms and then Login, the check is removed again. I guess this is because we've left the page for the login and do a complete rerender when we come back without having stored the state in a cookie. So this is fine from my point of view.

PR #156, opened 2023-03-30
Add ability to pass custom env vars to deployments through Session CR

### LINKED ISSUE #151: Add ability to add env vars to Session CRs
Add the ability to add env vars to IDE deployments either directly using a map, and/or by referencing ConfigMap(s).

Please assign me to the issue as I have started working on this. Creating an issue for discussion of the feature.

askarbozcan:
Closes #151 

**This PR adds following:**

* Changes to SessionSpec.java to accept
  1. Envvars directly through a map
  2. Envvars by passing in CM/Secret names as list
 
* Implementations of util functions to add environment variables to Deployment's 'env' and 'envFrom' maps
* Calling said util functions from `LazySessionHandler.createAndApplyDeployment`.

 
 _(Note that it is impossible to add this functionality to Eager handler, barring direct injection of env vars into running pods or env vars in App Definition)_

Non-PR related changes which had to be added:
* QoL script for building a local Docker image containing local maven repository
* a 'local' Dockerfile which uses the said local image to prevent re-downloading maven dependencies on every build
* **If monitor is disabled**, replaces placeholder string in template deployment 'placeholder-monitor-port' to integer '12345' to prevent parsing errors in `K8sUtil.loadAndCreateTypeWithOwnerReference`. The 'monitor' containerPort of Theia deployment is then removed in the `additionalModifications` Consumer of `K8sUtil.loadAndCreateTypeWithOwnerReference`.

[Session CRD is also updated to reflect the changes.](https://github.com/eclipsesource/theia-cloud-helm/pull/19)

**Not added:**
~* Support for Session CR env vars in the Theia service~
(Added in later commits)


askarbozcan:
Example Session CR:
```yaml
apiVersion: theia.cloud/v4beta
kind: Session
metadata:
  creationTimestamp: "2023-03-28T12:16:59Z"
  generation: 1
  name: finalcheck
  namespace: theiacloud
  resourceVersion: "5064dd94"
  uid: finalcheck
spec:
  appDefinition: theia-cloud-demo
  name: finalcheck
  user: finalcheck@theia.cloud
  envVars:
    kek: "visneli"
    hippo: "500kg"
  envVarsFromConfigMaps:
    - "my-config"
    - "my-other-config"
  envVarsFromSecrets:
    - "ssshh"

  ```

File: dockerfiles/operator/Dockerfile.local [hunk 1]
```diff
@@ -0,0 +1,24 @@
+FROM theia-cloud-local-maven-deps AS localDependencies
+
+
+FROM maven:3.8.4-openjdk-11 AS builder
+WORKDIR /root/.m2
+COPY --from=localDependencies /root/.m2 ./
+WORKDIR /operator
+COPY java/common ./common
```
lucas-koehler, line 8: Before this, just copy the relevant  pom.xml files for common and then run `mvn dependency:resolve` for them. this should only download the dependencies.
Afterwards, repeat the same for the operator.
Then, copy the rest of the folders and install like from line 10 onwards. With this, docker should only need to re-download dependencies when the pom files change.
You might need to build common before resolving dependencies for the operator

lucas-koehler (changes requested):
I like the idea of having the possibility to directly inject environment variables, load them from config maps and load secrets :)

I have some suggestions for improvement. I'll just list all of them. If it is too much effort for you to do all of that we can still discuss leaving some things out :)

1) The code is formatted differently. The code formatting rules are applied via Eclipse IDE configurations. Thus, it would be beneficial if you could re-format the code with that.

2) Regarding the local docker image: I like the idea of having faster builds in case you want to iterate often. However, copying the whole .m2 repository from the user home seems overkill. This will copy ALL downloaded maven bundles from your machine.
Instead, one could design a local Dockerfile differently to first load all maven dependencies and then copy the java code and build it. This way, the downloaded maven dependencies are cached by docker. I'll add some comments in the Dockerfile.local how this could potentially be achieved.

3) Rergarding the monitor port. I agree that we need to handle the monitor being disabled. However, rather than adding all this code to get rid of the port again, I think it would be easier to remove the monitor port from the templates `java/operator/org.eclipse.theia.cloud.operator/src/main/resources/templateDeployment.yaml` and `java/operator/org.eclipse.theia.cloud.operator/src/main/resources/templateDeploymentWithoutOAuthProxy.yaml`.
Instead, the port could just be added only if the monitor is enabled. If you can't spend the effort to rework this, it could also be possible to use your fix temporarily and tackle this later. @jfaltermeier Do you have an opinion on this?

jfaltermeier:
> 3. Rergarding the monitor port. I agree that we need to handle the monitor being disabled. However, rather than adding all this code to get rid of the port again, I think it would be easier to remove the monitor port from the templates `java/operator/org.eclipse.theia.cloud.operator/src/main/resources/templateDeployment.yaml` and `java/operator/org.eclipse.theia.cloud.operator/src/main/resources/templateDeploymentWithoutOAuthProxy.yaml`.
>    Instead, the port could just be added only if the monitor is enabled. If you can't spend the effort to rework this, it could also be possible to use your fix temporarily and tackle this later. @jfaltermeier Do you have an opinion on this?

So we usually start with loading the template-yaml as a string and replace the placeholders within this string. 
This string is then passed to the Kubernetes Java Client API to create the Java Object representation of the template. 
We then may do some further changes to the Java object using Java code. 
Only this object is then passed to Kubernetes again to actually create the Kubernetes resource. 

So we could remove this port from the template itself and add it in the Java-Code instead if required. 
We already do programatic changes to add volume claims, image pull secrets, bandwidth limits in `LazySessionHandler.createAndApplyDeployment`

I think @sgraband is currently working on fixing this. So I agree with Lucas that it is fine to use your fix temporarily and then we either rebase this change on the fixed version or remove the workaround again.

sgraband:
Yes i am working on a fix. I am completely fine with rebasing and later on replacing the workaround with the solution i am working on (It will also fix some additional minor issues that might happen).

dockerfiles/operator/Dockerfile.local
askarbozcan: I have made some changes, however no matter what I tried, it still downloads in build stage (not dependency resolving stage). Do you want to try developing such a Dockerfile that won't re-download all dependencies (unless pom changed), because I have failed. 

I tried doing something similar with maven multi-stage dockerfiles in another project, but it failed there as well.

Although heavy-handed somewhat this approach of just copying all dependencies from a local image of local maven repository is the only one that allowed me to have sub ~20s build times.

lucas-koehler: Unfortunately, I currently do not have the time to investigate such a Dockerfile, sorry :( . Do I understand you correctly that the adapted Dockerfile in your commit https://github.com/eclipsesource/theia-cloud/pull/156/commits/58ddd634d5eaa8e44ad58d60869d0ce864dd0aa9 did not solve your problem?

askarbozcan: Yes, it does not solve it.

However I found another way: 

RUN **--mount=type=cache,target=/root/.m2** \
    cd /operator/common/maven-conf && \

This caches the dependencies, and only pulls them if missing/first build. I propose making this the main image, as cache can be turned off with --no-cache

askarbozcan:
Service extension done. Tested it on a cluster, it works. Anything I've missed?
Left to do:
* Unit tests
* (Probably should be in another PR) Caching of custom envs in Workspace CR
* Squash all commits of the PR into one
* Remove some unnecessary files from the corresponding Helm PR

jfaltermeier:
FYI, I will have a look at this PR this week, either on Thursday or Friday.

jfaltermeier (changes requested):
Thank you very much, I have a few smaller comments, which I've added on the files. 

Besides that I currently pushed your changes here: https://github.com/eclipsesource/theia-cloud/commits/feature/sessioncr_envvar_proper and added three additional commits. One formats the code, the other two regenerate the OpenAPI spec and the typescript code for the service changes (see https://github.com/eclipsesource/theia-cloud/blob/main/doc/docs/OpenAPI.md)

Feel free to cherry-pick these changes to your branch. 

I mainly tested with fromMap environment variables which worked as expected. I also tested the service shortly with postman and this example workload `{"appDefinition":"theia-cloud-demo","user":"bar@theia-cloud.io","appId":"asdfghjkl","ephemeral":true,"env": {"fromMap":{"Key2":"Value2","Key1":"Value1","Key3":"Value3"}}}` which worked as well. 

So after addressing the smaller comments above, we may merge this from my point of view. 

Regarding the process for merging I would suggest that we merge it after #158 , since this includes the monitor fix mentioned above. We would then take care of resolving the merge conflict in this case, so there is nothing to do from your side.

dockerfiles/operator/Dockerfile.withcache
```diff
@@ -0,0 +1,20 @@
+FROM maven:3.8.4-openjdk-11 AS builder
+WORKDIR /operator
+COPY java/common ./common
+COPY java/operator ./operator
+RUN --mount=type=cache,target=/root/.m2 \
```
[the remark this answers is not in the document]
askarbozcan: I cannot test this because I have no access to a Windows machine.

As far as I know, though, cache mount is similar to cached layers, meaning the caching is handled by Docker itself and the "target" dir is a kind of identifier for the cache. It has nothing to do with host's filesystem.

Barring all of that, Docker on non-Linux systems runs on Linux VMs anyway so IMO there should be no issues. I do agree, however, that this needs to be tested on a Windows machine just to be sure.

askarbozcan:
The requested changes have been added. Thanks for your review!

askarbozcan:
Tested again (on my local cluster).
Works fine for all the ways of inputting env vars (directly as a map, from CM, from Secrets)

PR #157, opened 2023-03-30
Support zero max instances for app definitions

lucas-koehler:
- Setting a app definitions `maxInstances` to `0` now prevents any launch
- Negative or omitted `maxInstances` allows indefinite launches

Further small changes:
- Disable printing transfer progress in operator and service docker builds to avoid log pollution
- Reduce log level for app definitions allowing indefinite instances. The reason is that the indefinite instances are the expected result directly following from the app definition. It seems unnecessary to log this as info over and over again.


jfaltermeier (approved):
Thanks, lgtm and works as expected!

PR #160, opened 2023-04-11
trynow: Update wording regarding login

jfaltermeier:


node/try-now-page/src/components/LoginInfo.tsx
```diff
@@ -3,12 +3,11 @@ export const LoginInfo: React.FC = () => (
...
+    <pre>We only need your login data for preventing misuse.</pre>
     <pre>
-        We only need your login data for technical reasons (e.g. avoiding crypto-mining or other harmful actions).
-    </pre>
-    <pre>We will not contact you via your e-mail or use it for any other marketing purposes.</pre>{' '}
-    <pre>
-        You can login using your Github account (You will be forwarded to our Keycloak instance).
-    </pre>
+      We <strong>never</strong> contact you via your e-mail and will <strong>never</strong> use it for any other
+      marketing purposes.
```
lucas-koehler, line 9: As this is in a `<pre>`, there always is a line break between other and marketing, isn't it?
While not introduced with this PR, it seems a little strange to me that we use `pre` tags at all.
Is there a specific reason for this. Otherwise, it seems more sensible to use paragraphs (`<p>`).

lucas-koehler (commented):
Wording LGTM me. I noticed that it is a little strange to use `<pre>` tags.

node/try-now-page/src/components/LoginInfo.tsx
jfaltermeier: I've replaced it with `p`

lucas-koehler (approved):
LGTM now, thanks. Note that I did not test this as the effort seemed overkill for such a straight forward change ;)

Issue #69, opened 2022-08-19
Init operations for workspaces


jfaltermeier:
We might offer operations that are performed when a new workspace is created, for example cloning a given git repository. 
Initial ideas would be either by an init-container or by a Theia-Extension. 
However we should check for other use cases as well and how this may be integrated in an extensible way.

jfaltermeier:
Keep open

jfaltermeier:
Keep open

sgraband:
Keep open

jfaltermeier:
Keep open

PR #182, opened 2023-05-22
Init operations for workspaces with Git Init as initial implemetation #69

jfaltermeier:
CRD, RBAC changes: https://github.com/eclipsesource/theia-cloud-helm/pull/26/files

* add a new docker image (`theiacloud/theia-cloud-git-init`) that can be used to clone a git repository and checkout a specific pointer
  * support HTTPS and SSH auth
  * can be tested outside of Theia Cloud with the instructions in `python/git-init/README.md` (you may have to create/find approriate Git repositories to test though. Note: https://gitlab.eclipse.org allows to create private forks of existing projects in your user namespace!)
* arguments to create a init container using the new image are passed via the new `initOperations` array of a session. The container is created in `GitInitOperationHandler`
  * for private repositories the handler currently expects that the secret is available as a kubernetes secret already
  * besides that the secret has to be annotated to be used by the operator. Otherwise it will not be exposed. See the sample secrets in `python/git-init/README.md`
  * since the secrets are only available in the init container, they cannot be read from the main IDE
* Via the `InitOperationHandler` new init operation handlers may be added to the Operator
* extends the REST Service API and Typescript Common API to allow creating sessions with init operations

Open issues / Known limitations are:
* the actual IDE has no git config with the user email yet
* If we did a clone using SSH the private key is not made available to the main IDE yet
* Does not work with ephemeral sessions, because we have no volume we could mount in the init container
* We have no API to create the required Kubernetes Secrets. Currently it is expected that an admin user adds the credentials to kubernetes with the required annotations

For the credential handling I am not sure if this should be a first level Theia Cloud Feature. We could offer some library/util that helps with creating the secrets/adding the annotations, but I think this should be used from client code, e.g. from a dashboard implementation.


java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/impl/GitInitOperationHandler.java
```diff
@@ -0,0 +1,272 @@
...
+	    String[] split = repository.toLowerCase().split(Pattern.quote("://"), 2);
+	    if (split.length != 2) {
+		LOGGER.error(LogMessageUtil.formatLogMessage(correlationId, MessageFormat
+			.format("Failed to check whether repository {0} contains any user information. ", repository)));
+		return false;
+	    }
+	    String repositoryWithoutProtocol = split[1];
+	    if (repositoryWithoutProtocol.contains("@")) {
+		if (repositoryWithoutProtocol.split(Pattern.quote("@"))[0].contains(":")) {
+		    /* username and password part of URL */
```
lucas-koehler, line 178: Is this if empty on purpose?

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/impl/GitInitOperationHandler.java
```diff
@@ -0,0 +1,272 @@
...
+		    injectPassword = true;
+		}
+	    } else {
+		injectUsername = true;
+		injectPassword = true;
+	    }
+	}
+
+	LOGGER.info(LogMessageUtil.formatLogMessage(correlationId,
+		MessageFormat.format("Inject username: {0}; Inject password: {1}", injectUsername, injectPassword)));
```
lucas-koehler, line 190: Should we log credentials? This might expose secrets in the log.

java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/BaseResource.java
```diff
@@ -41,6 +41,14 @@ protected String evaluateRequest(ServiceRequest request) {
 	return correlationId;
     }
 
+    protected boolean isSameAuthenticatedUser(TheiaCloudUser user, String requestUser) {
+	if (user.isAnonymous()) {
+	    /* if authentication is disabled we trust the user information */
```
lucas-koehler, line 46: I think the safer check adhering to the comment is to check that keycloak is disabled by querying the application properties.

lucas-koehler (commented):
Initial code review. I will test this later.
Code structure and code looks mostly good to me already :) . I added some inline comments.

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/k8s/resource/SessionSpec.java
```diff
@@ -217,37 +277,82 @@ public boolean equals(Object obj) {
...
 	return true;
     }
 
     @Override
     public String toString() {
 	return "SessionSpec [name=" + name + ", appDefinition=" + appDefinition + ", user=" + user + ", url=" + url
-		+ ", error=" + error + ", workspace=" + workspace + ", lastActivity=" + lastActivity + "]";
+		+ ", error=" + error + ", workspace=" + workspace + ", lastActivity=" + lastActivity
+		+ ", sessionSecret=" + sessionSecret + ", envVars=" + envVars + ", envVarsFromConfigMaps="
+		+ envVarsFromConfigMaps + ", envVarsFromSecrets=" + envVarsFromSecrets + ", initOperations="
```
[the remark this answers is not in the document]
jfaltermeier: Those are just the names of the secrets under which they are stored in k8s and not their actual values

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/impl/GitInitOperationHandler.java
jfaltermeier: This just a boolean whether they will be injected or not

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/k8s/resource/SessionSpec.java
```diff
@@ -217,37 +277,82 @@ public boolean equals(Object obj) {
...
+		+ initOperations + "]";
     }
 
     public static boolean isEphemeral(String workspace) {
 	return workspace == null || workspace.isBlank();
     }
 
+    public static class InitOperation {
+
+	@JsonProperty("id")
```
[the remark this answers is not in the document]
jfaltermeier: I've added id as required in the CRD: https://github.com/eclipsesource/theia-cloud-helm/blob/fbbceb8196189367e76106504a03c2f02bfc737e/charts/theia-cloud-crds/templates/session-spec-resource.yaml#L70C25-L70C25

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/impl/GitInitOperationHandler.java
```diff
@@ -0,0 +1,272 @@
...
+
+	if (isHTTP(repository)) {
+	    if (!injectHTTPRepoCredentials(correlationId, secret, secretName, repository, gitInitContainer)) {
+		return;
+	    }
+	} else {
+	    if (!injectSSHRepoCredentials(correlationId, secret, secretName, repository, gitInitContainer, volumes)) {
+		return;
+	    }
+	}
```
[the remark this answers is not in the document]
jfaltermeier: All the logging is done in the inject methods itself. 
The returns are executed when there was a problem. 
Reason is that if we add more init code at the end at some point (e.g. injecting the SSH Key in the IDE) or if an adopter overrides the handler and calls `super`, we only run this code if there were no issues. 
I've added more comments, because it looks weird indeed.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/impl/GitInitOperationHandler.java
jfaltermeier: Yes, `injectUsername` and `injectPassword` are false by default and we won't inject either if both username and password are part of the repo url. 
I've adopted the comment

lucas-koehler:
keep open

michaelsharpe:
Any word on this?  This would be amazing for some work we are looking to use Theia Cloud for.

jfaltermeier:
> Any word on this? This would be amazing for some work we are looking to use Theia Cloud for.

Thank you for your interest in this feature! Currently, our main priority is transferring the codebase to the Eclipse Foundation. Once this process is complete, we plan to revisit and continue development on earlier work, including this ticket.

PR #186, opened 2023-06-14
Limit user access to their own resources

### LINKED ISSUE #180: Limit user access to their own resources

This only affects deployments using user authentication.

Currently, users can list and launch session of workspaces of all users. With this, user can get information about other users.

Users must only be able to launch, see, edit and delete their own resources (i.e. workspaces, sessions).

There is no alternative. Users must not be able to access other users' resources.

### Cluster provider

_No response_

### Additional information

_No response_

lucas-koehler:
Fix #180 

## Summary
This PR extends the REST service to limit user access to their own resources for all calls.
Now, Users may only create, start, stop, list, and get performance metrics for their own sessions and workspaces. Trying to do any of this in the name of another user (by specifying them in the request's user field) throws a `403` Forbidden HTTP Error.

Furthermore, all authenticated methods (everything but ping and and Session.activity) except for the launch method, can no longer be used in anonymous mode. This is achieved via a new `@NoAnonymousAccess` annotation triggering an interceptor. With this, anonymous requests return a `403` Forbidden HTTP Error in anonymous mode.

## How to test

- Use the terraform testing config
- Execute the minikube setup
- Build service in Minikube by first pointing docker to minikube's docker registry via `eval $(minikube docker-env)`
- In `terraform/test-configurations/2_try-now/theia_cloud.tf`, adapt the helm values to use the custom built service image by adding config to set the helm values accordingly, e.g. like this:
```terraform
# Only pull missing images. This is needed to use images built locally in Minikube
  set {
  name  = "imagePullPolicy"
  value = "IfNotPresent"
  }

  set {
    name  = "service.image"
    value = "theia-cloud-service:180-user-resource-access"
  }
```
- Install the theia-cloud-base and try-now setups
- Use the node testing page to execute rest calls on the service. Adapt the service url and keycloak url in `node/testing-page/src/App.tsx`

## Implementation details

<details>

### Generalize user scoped requests in new class 
Add a new sub class UserScopedServiceRequest of ServiceRequest to unify
specification of a user for all requests using one.

### Allow user-scoped requests only for the same authenticated user 

So far, most methods did not verify that the user named in the requestis the same as the authenticated user.

This commit fixes this by overloading the evaluateRequest method in BaseResource with a second version taking a UserScopedServiceRequest.

For these requests, request.user must either be the same as the authenticated user's identifier or be omitted. In the latter case, the user's identifier is used automatically.

In anonymous mode, the request.user property must be set because there is no other way to determine the user.

In both cases, the resolved user identifier to use is returned as part of the new type `EvaluatedRequest`.

### Add annotation to prevent any anonymous access 
Adds a new interceptor binding annotation `@NoAnonymousAccess` and the corresponding interceptor implementation that blocks all anonymous access to a resource class or method.

### Block any anonymous access to various session and workspace methods 

- Block any anonymous access for methods:
  - Session: list, performance, start, stop
  - Workspace: create, delete, list
- Add new error code 474 for invalid session name
- Adds tests:
  - Session: performance, list, start, verify access control annotations
  - Workspace: create, list, verify access control annotations
  - Root: Ensure users cannot launch sessions for other users
 
### Misc
- Extend testing page with ping and launch buttons
- Add imagePullPolicy hint to testing terraform config
- Disable mvn transfer progress for service test ci
</details>

### DIFF
--- a/.github/workflows/service-ci.yml
+++ b/.github/workflows/service-ci.yml
@@ -19,10 +19,10 @@ jobs:
         distribution: 'temurin'
     - name: Build TheiaCloud common
       run: |
-        mvn clean install -f java/common/maven-conf
-        mvn clean install -f java/common/org.eclipse.theia.cloud.common
+        mvn clean install -f java/common/maven-conf --no-transfer-progress
+        mvn clean install -f java/common/org.eclipse.theia.cloud.common --no-transfer-progress
     - name: Run TheiaCloud service tests
-      run: mvn clean verify -fae -f java/service/org.eclipse.theia.cloud.service
+      run: mvn clean verify -fae -f java/service/org.eclipse.theia.cloud.service --no-transfer-progress
 
   docker:
     needs: tests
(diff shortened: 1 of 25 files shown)


sgraband (approved):
Thank you! The changes look good and work great! This is definitely a huge improvement :+1:

PR #194, opened 2023-07-11
Operator crashes in eager start mode if monitor is not used

paulreuter:
Fixes NullPointerException if monitor is not used, and yaml parse error if monitor is in use


sgraband (commented):
Thank you for the changes @paulreuter! I definately think it makes sense that the monitor does also work on eager Pods. 
Just one minor thing: Could we extract the duplicated code to a seperate method to reuse it?

sgraband (approved):
Thanks a lot! LGTM :+1:

PR #196, opened 2023-07-11
Eager start can not retrieve existing services

### LINKED ISSUE #195: Eager start can not retrieve existing services
### Describe the bug

There is a discrepancy in the read/write of the naming schemes for services and pods.

NamingUtil.asValidName enforces that object names start with an alphabetic char by prepending 'a'

TheiaCloudK8sUtil.extractIdFromName expects a pure numerical segment and throws a NumberFormatException due to the prefix

The operator then crashes by trying to create an already existing service.

### Expected behavior

Existing services can be retrieved

### Cluster provider

GKE

### Version

0.8.0-MS10v1

### Additional information

_No response_

paulreuter:
Fixes #195

### DIFF
--- a/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/util/TheiaCloudK8sUtil.java
+++ b/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/util/TheiaCloudK8sUtil.java
@@ -69,6 +69,7 @@ public final class TheiaCloudK8sUtil {
 	String name = metadata.getName();
 	String[] split = name.split("-");
 	String instance = split.length == 0 ? "" : split[0];
+	// kubernetes names must not start with letter, remove automatically added prefix
 	instance = instance.length() == 0 ? "" : instance.charAt(0) == NamingUtil.VALID_NAME_PREFIX ? instance.substring(1) : instance; 
 	return instance;
     }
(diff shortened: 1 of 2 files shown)


java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/util/TheiaCloudK8sUtil.java
```diff
@@ -68,6 +68,7 @@ public static String extractIdFromName(ObjectMeta metadata) {
 	String name = metadata.getName();
 	String[] split = name.split("-");
 	String instance = split.length == 0 ? "" : split[0];
+	instance = instance.length() == 0 ? "" : instance.charAt(0) == 'a' ? instance.substring(1) : instance;
```
lucas-koehler, line 71: Could you add a comment why this is necessary, please? Otherwise, this might be very confusing when looking at this again in the future.
Even better would be adding a constant with the prefix `a` to `NamingUtil` and getting it from there :)

lucas-koehler (changes requested):
Hi @paulreuter ,
thanks for the contribution :heart: . I added a small comment inline.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/util/TheiaCloudK8sUtil.java
paulreuter: Alright, that should do it.

lucas-koehler (approved):
LGTM now, thanks again :)

Issue #32, opened 2022-05-12
Watch did stop on Test Cluster


jfaltermeier:
The watch for workspace events has stopped on the test cluster. There was no exception logged. 

Our manual timer based watch logged some exceptions (see below), so it might have something to do with the connection to the cluster API. 

We should try to add more logging to figure out how to detect this state, so that we may set up the watches again. 

```
io.fabric8.kubernetes.client.KubernetesClientException: Operation: [list]  for kind: [TemplateSpecResource]  with name: [null]  in namespace: [theiacloud]  failed.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:130) ~[operator-0.7.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.listRequestHelper(BaseOperation.java:138) ~[operator-0.7.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:413) ~[operator-0.7.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:402) ~[operator-0.7.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.list(BaseOperation.java:83) ~[operator-0.7.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at org.eclipse.theia.cloud.operator.KillAfterRunnable.run(KillAfterRunnable.java:60) [operator-0.7.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) [?:?]
	at java.util.concurrent.FutureTask.runAndReset(Unknown Source) [?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) [?:?]
	at java.lang.Thread.run(Unknown Source) [?:?]
Caused by: java.net.ConnectException: Failed to connect to /10.52.0.1:443
[...]

PR #214, opened 2023-08-01
Detection for stuck kubernetes watchers #210

### LINKED ISSUE #210: [Operator] Detect stuck operator

Once we allow multiple replicas of the operator we should make sure that the probes are implemented. 

* We don't offer an API endpoint, so implementing the kubernetes health/ready/live probes does not really make sense. 
* We should still try to detect those cause, but then simply stop our java application. 
* What cases do we have where we don't crash but are stuck
  * when watches for CRs are not called anymore
    * we had issues with this because of a kubernetes bug here: https://github.com/eclipsesource/theia-cloud/issues/32
    * I don't think we have a way to detect issues with Watches fast, meaning that we would have to detect it based on some timeframe (e.g. if we haven't received an event or have been asked if we can reconnect within two hours, we assume that watch to be dead)
  * Look for calls with active waiting that might get stuck
    * don't do any active waiting on the main thread
    * all waits must have a timeout
      * [x] ResourceClient.watchUntil
      * [x] AddedHandlerUtil.updateSessionURLAsync
    * Calls to the kubernetes-client will either throw an exception or return immediately, so there should be no action required atm.  

Stop the operator if it gets stuck. 

-

### Cluster provider

_No response_

### Additional information

_No response_

jfaltermeier:
* extend our watchers to remember when they were last used
* add configurable timeout value where we stop the operator when watchers appear stuck
* Historically we had this issue once because of a kubernetes bug: https://github.com/eclipsesource/theia-cloud/issues/32

closes #210

### DIFF
--- a/java/operator/org.eclipse.theia.cloud.operator/log4j2.xml
+++ b/java/operator/org.eclipse.theia.cloud.operator/log4j2.xml
@@ -15,6 +15,10 @@
 			additivity="false">
 			<AppenderRef ref="Console" />
 		</Logger>
+		<Logger name="org.eclipse.theia.cloud.operator.TheiaCloudImpl" level="info"
+			additivity="false">
+			<AppenderRef ref="Console" />
+		</Logger>
 		<Logger name="org.eclipse.theia.cloud.operator" level="trace"
 			additivity="false">
 			<AppenderRef ref="Console" />
--- a/demo/k8s/appdefinitions/theia.yaml
+++ b/demo/k8s/appdefinitions/theia.yaml
@@ -5,7 +5,7 @@ metadata:
   namespace: theiacloud
 spec:
   downlinkLimit: 30000
-  image: theiacloud/theia-cloud-demo:0.8.1.OSWeek23-v1
+  image: theiacloud/theia-cloud-demo:0.8.1.OSWeek23-v2
   imagePullPolicy: IfNotPresent
   ingressname: theia-cloud-demo-ws-ingress
   limitsCpu: "2"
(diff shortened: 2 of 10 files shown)


java/operator/org.eclipse.theia.cloud.operator/log4j2.xml
```diff
@@ -15,6 +15,10 @@
 			additivity="false">
 			<AppenderRef ref="Console" />
 		</Logger>
+		<Logger name="org.eclipse.theia.cloud.operator.TheiaCloudImpl" level="info"
+			additivity="false">
+			<AppenderRef ref="Console" />
+		</Logger>
```
sgraband, line 21: Is this intended or an oversight from debugging?

sgraband (approved):
Thank you! looks good to me! :+1: Just one minor question below. If intended, feel free to merge.

java/operator/org.eclipse.theia.cloud.operator/log4j2.xml
jfaltermeier: I've added this because of this log: https://github.com/eclipsesource/theia-cloud/blob/54c83d552e1f0539a51c6a8ec5cc09182c692d4b/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/TheiaCloudImpl.java#L254
This would get spamed to the log every minute as the default log level for the `org.eclipse.theia.cloud.operator` package is `trace` at the moment.

sgraband: Thanks for the clarification :+1:

PR #223, opened 2023-08-02
Add validation in service

xai:
* Increment kubernetes client update for x-kubernetes-validation rules support
* Add check for matching app definition


jfaltermeier (approved):
Thanks, lgtm and works as expected. 
I've tested with starting a non-ephemeral sessions to create a workspace and then stopped the session again. Then I manipulated the Workspace CR and changed the stored app definition. When attempting to start a session again, this gets denied as expected.

jfaltermeier:
FYI: for further testing I rebased this branch and added commit that updates the CRD version in the java code https://github.com/eclipsesource/theia-cloud/tree/issues/199-jf

PR #225, opened 2023-08-07
Eager start Ingress rule calculates bad path

### LINKED ISSUE #197: Eager start Ingress rule calculates bad path
### Describe the bug

IngressPathProviderImpl used by eager start returns a non-usable url fragment (full object toString):
getBasePath() + appDefinition.getSpec() + "-" + instance

This causes a "pcre_compile() failed: range out of order in character class" in the nginx rewrite due to the dashes being interpreted as regex ranges

The intention is probably to use appDefinition.getSpec().getName() instead. However, I think using the sessions randomly generated uid as in the lazy start might be better. What do you think?

### Expected behavior

Eager start should properly launch sessions

### Cluster provider

_No response_

### Version

_No response_

### Additional information

_No response_

paulreuter:
fixes #197


jfaltermeier (approved):
Thanks, LGTM! I would also be fine with switching it to use the app definition's uid + instance instead of the name, as suggested on the ticket. So depending on your preference we may merge this or you could update the PR. 

(In general we haven't used/tested the eager start lately, because we don't have a solution for attaching persisted storage to already started pods yet. This requires the pod to restart, so we would need some kind of network drive for the eager start.)

PR #227, opened 2023-08-10
Fix Monitor service

sgraband:
- Skip reverse proxy for monitor requests 

Add `--skip-auth-route` option to oauth2-proxy container.
Ignore all requests starting with `/monitor`.

- Adjust monitor-vscode terraform config

Fix to work with authentication.
Adjust values to work with subdomains.

- Only display one timeout message at a time 

Before a new message would have been created with each ping.
If a message is already displayed we do not display another one.

To test:
Run the terraform configs with `2-02_monitor-vscode` with locally built operator from this PR.
The vscode and theia extension can be tested locally to avoid having to bundle them in a docker image.


node/monitor/src/modules/activity-tracker-module.ts
```diff
@@ -84,20 +86,23 @@ export class ActivityTrackerModule implements MonitorModule {
...
 
-    message.then(answer => {
-      if (answer === yesOption) {
-        this.reportActivity('Popup dialog was confirmed');
-      }
-    });
+      message.then(answer => {
+        if (answer === yesOption) {
+          this.reportActivity('Popup dialog was confirmed');
+        }
```
jfaltermeier, line 103: Do we need an else case to reset `this.messageAlreadyDisplayed` just in case? I think if we get an answer, then the dialog is not showing anymore, so we might have to make sure a new one can be shown.

jfaltermeier (commented):
Looks good to me. I have added one comment inline. 
I've also added a commit which contains some changes to the terraform charts/build instructions I used for testing.

node/monitor/src/modules/activity-tracker-module.ts
sgraband: Good idea! I added it in a commit.

jfaltermeier (approved):

PR #238, opened 2023-09-07
Fix issue terraform scripts not working on windows

dmm9:
Theia-cloud does not work on Windows out-of-the-box. Probably due to the fact that the local installation of terraform uses cmd.exe under the hoods. There are 2 issues:
- `--type='json'` --> Single or double quotes cause an error. Removing the quotes works on windows.
- `-p '[{ \"...` --> Also the initial quotes and escaped quotes cause issues. I tried several combinations with no success. The only way I have found is to extract the patch to a local variable.

I didn't test in linux, I hope the CI tests it. Otherwise please test with linux before merging.

### DIFF
--- a/terraform/modules/helm/main.tf
+++ b/terraform/modules/helm/main.tf
@@ -157,6 +157,10 @@ resource "kubectl_manifest" "selfsigned_issuer" {
 }
 
 locals {
+  # local_exec_quotes is a helper function to deal with different handling of
+  # quotes between linux and windows. On linux, it will output "'". On windows,
+  # it will output "".
+  local_exec_quotes = startswith(abspath(path.module), "/") ? "'" : ""
   jsonpatch = jsonencode([{
       "op" = "add",
       "path" = "/spec/template/spec/containers/0/args/-",
@@ -215,7 +219,7 @@ resource "helm_release" "keycloak" {
   # Below command connects to the cluster in the local environment and patches the ingress-controller accordingly. 
   # Theia Cloud is then installed with path based hosts reusing the same certificate. 
   provisioner "local-exec" {
-    command = "kubectl patch deploy ingress-nginx-controller --type=json -n ingress-nginx -p ${local.jsonpatch} && kubectl wait pods -n ingress-nginx -l app.kubernetes.io/component=controller --for condition=Ready --timeout=90s && kubectl wait certificate -n keycloak ${var.hostname}-tls --for condition=Ready --timeout=90s"
+    command = "kubectl patch deploy ingress-nginx-controller --type=${local.local_exec_quotes}json${local.local_exec_quotes} -n ingress-nginx -p ${local.local_exec_quotes}${local.jsonpatch}${local.local_exec_quotes} && kubectl wait pods -n ingress-nginx -l app.kubernetes.io/component=controller --for condition=Ready --timeout=90s && kubectl wait certificate -n keycloak ${var.hostname}-tls --for condition=Ready --timeout=90s"
   }
 }
 
--- a/terraform/modules/helm/main.tf
+++ b/terraform/modules/helm/main.tf
@@ -156,6 +156,14 @@ resource "kubectl_manifest" "selfsigned_issuer" {
   yaml_body  = file("${path.module}/clusterissuer-selfsigned.yaml")
 }
 
+locals {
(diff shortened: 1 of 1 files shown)


terraform/modules/helm/main.tf
```diff
@@ -156,6 +156,14 @@ resource "kubectl_manifest" "selfsigned_issuer" {
   yaml_body  = file("${path.module}/clusterissuer-selfsigned.yaml")
 }
 
+locals {
+  jsonpatch = jsonencode([{
```
jfaltermeier, line 160:
```suggestion
  local_exec_quotes = startswith(abspath(path.module), "/") ? "'" : ""
  jsonpatch = jsonencode([{
```

terraform/modules/helm/main.tf
```diff
@@ -207,7 +215,7 @@ resource "helm_release" "keycloak" {
   # Below command connects to the cluster in the local environment and patches the ingress-controller accordingly. 
   # Theia Cloud is then installed with path based hosts reusing the same certificate. 
   provisioner "local-exec" {
-    command = "kubectl patch deploy ingress-nginx-controller --type='json' -n ingress-nginx -p '[{ \"op\": \"add\", \"path\": \"/spec/template/spec/containers/0/args/-\", \"value\": \"--default-ssl-certificate=keycloak/${var.hostname}-tls\" }]' && kubectl wait pods -n ingress-nginx -l app.kubernetes.io/component=controller --for condition=Ready --timeout=90s && kubectl wait certificate -n keycloak ${var.hostname}-tls --for condition=Ready --timeout=90s"
+    command = "kubectl patch deploy ingress-nginx-controller --type=json -n ingress-nginx -p ${local.jsonpatch} && kubectl wait pods -n ingress-nginx -l app.kubernetes.io/component=controller --for condition=Ready --timeout=90s && kubectl wait certificate -n keycloak ${var.hostname}-tls --for condition=Ready --timeout=90s"
```
jfaltermeier, line 218:
```suggestion
    command = "kubectl patch deploy ingress-nginx-controller --type=${local.local_exec_quotes}json${local.local_exec_quotes} -n ingress-nginx -p ${local.local_exec_quotes}${local.jsonpatch}${local.local_exec_quotes} && kubectl wait pods -n ingress-nginx -l app.kubernetes.io/component=controller --for condition=Ready --timeout=90s && kubectl wait certificate -n keycloak ${var.hostname}-tls --for condition=Ready --timeout=90s"
```

jfaltermeier (changes requested):
Hi, thank you very much for the contribution. Unfortunately the `sh` on linux behaves different than `cmd` on windows, so the changes don't work on my system. 

I've made a suggestion where we check whether the file path starts with `/`. If so we add the single quotes, otherwise we don't use quotes. I would guess that a windows path will start with something like `C:`. Could you please test if this works for you as well?

terraform/modules/helm/main.tf
dmm9: Yes it works on windows. But this function kind of obfuscates the logic. I would leave a comment for the next ones reading it.
```suggestion
  # local_exec_quotes is a helper function to deal with different handling of
  # quotes between linux and windows. On linux, it will output "'". On windows,
  # it will output "".
  local_exec_quotes = startswith(abspath(path.module), "/") ? "'" : ""
  jsonpatch = jsonencode([{
```

terraform/modules/helm/main.tf
dmm9: Yes it works also on windows. The produced command doesn't change. For reference, I leave here the console output on windows:
```ps
module.helm.helm_release.keycloak (local-exec): Executing: ["cmd" "/C" "kubectl patch deploy ingress-nginx-controller --type=json -n ingress-nginx -p [{\"op\":\"add\",\"path\":\"/spec/template/spec/containers/0/args/-\",\"value\":\"--default-ssl-certificate=keycloak/xx.xx.xx.xx.sslip.io-tls\"}] && kubectl wait pods -n ingress-nginx -l app.kubernetes.io/component=controller --for condition=Ready --timeout=90s && kubectl wait certificate -n keycloak xx.xx.xx.xx.sslip.io-tls --for condition=Ready --timeout=90s"]
```

jfaltermeier (approved):

PR #245, opened 2023-10-19
Make version of oauth2 proxy configurable via operator param

sgraband:
This avoids always using the latest image.
For example if there are breaking changes.
Defaults to latest.

Linked helm repo PR: https://github.com/eclipsesource/theia-cloud-helm/pull/44
To test build the operator, create a deployment using the helm chart from the link above and make sure to use keycloak.
When creating a session, the pod should contain the oauth2 proxy with the version you specified in the helm chart.

Note: This was originally brought up, because we had issues with the 7.5.0 version of the oauth2 proxy. I can however, no longer reproduce them. So testing any reasonably new version should work.

### DIFF
--- a/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/TheiaCloudArguments.java
+++ b/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/TheiaCloudArguments.java
@@ -218,7 +218,7 @@ public class TheiaCloudArguments {
 	return continueOnException;
     }
 
-    public String getoAuth2ProxyVersion() {
+    public String getOAuth2ProxyVersion() {
 	return oAuth2ProxyVersion;
     }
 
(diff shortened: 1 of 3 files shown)


java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/TheiaCloudArguments.java
```diff
@@ -118,6 +118,10 @@ public enum BandwidthLimiter {
 	    "--continueOnException" }, description = "Whether the operator will continue to run in case of unexpected exceptions.", required = false)
     private boolean continueOnException;
 
+    @Option(names = {
+	    "--oAuth2ProxyImage" }, description = "The version to use of the quay.io/oauth2-proxy/oauth2-proxy image.", required = false, defaultValue = "latest")
+    private String oAuth2ProxyImage;
```
lucas-koehler, line 123: Minor but I think it makes more sense to rename this to `oAuth2ProxyVersion` like in the helm chart to avoid confusion.

```suggestion
    private String oAuth2ProxyVersion;
```

lucas-koehler (commented):
I have one minor comment inline. Besides that it looks good to me

lucas-koehler (approved):
LGTM, thanks :)

PR #259, opened 2023-12-08
Rename publish npm script to avoid rerun of publish

sgraband:
_Small fix to fix failing builds: https://github.com/eclipsesource/theia-cloud/actions/runs/7140492012. Artifacts have been published correctly though. So no rerun is necessary._

Renamed publish script to publish:latest.
Before the script was executed everytime we ran npm publish. 
This then fails because you cannot publish the same version twice. 
Also prevents the latest tag to be applied on next publish. Also it probably would have ended in a loop.

### DIFF
--- a/.github/workflows/reusable-npm.yml
+++ b/.github/workflows/reusable-npm.yml
@@ -76,6 +76,6 @@ jobs:
         run: npm ci
 
       - name: Publish new version and latest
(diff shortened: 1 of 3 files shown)


sgraband:
@jfaltermeier Could you take another look, i also fixed the demo publishing. The changes are minor and i linked the successful runs where the demos were published.

jfaltermeier (approved):

PR #277, opened 2024-01-23
Revert "Regenerate common package"

sgraband:
With that commit the try-now page stopped working.

This commit will be backported and be part of the 0.9.1 release, as it breaks the application.
Once merged i will add it to the release branch.


node/landing-page/package.json
```diff
@@ -11,7 +11,7 @@
   "dependencies": {
     "@eclipse-theiacloud/common": "0.9.0",
     "@types/uuid": "^8.3.2",
-    "axios": "^1.6.5",
+    "axios": "^0.25.0",
```
jfaltermeier, line 14: should we use `^0.27.2` here as well?

sgraband: I will take a look at regenerating the code again, once the update is finished. So i would just keep it as it was before for now, if you do not mind.

jfaltermeier (approved):

PR #280, opened 2024-01-25
Fix workflows and monitor extension

sgraband:
**Update to 0.10.0-next**

**Do not build on full release main pushes**

Introduce step that checks the version so we can use the information in the IF statment of jobs.
Unfortunately ENVs are not available there, hence the workaround.
Queue monitor-vscode after theia-cloud-demo so they can be build together with one run.
Automate the detection of the version for the demo flow.
Now there is no need to specify on how they should be published.

**Fix bundling of monitor extension**

Remove `node/monitor` from the workspaces setup, because `vsce` is incompatible with it.
Still build the extension together with other packages.
Provide quick command to update the extension in the demo.
Update extension in demo.
Remove duplicate settings.
Fix linting setup and lint packages that were not linted.
Fix minor linting issues.
Make tsconfigs more clear.


node/monitor/package.json
jfaltermeier, line 1: I think this file should be formatted once to remove the tabs that were introduced, e.g. in front of the `webpack` script

jfaltermeier (commented):
Looks very good, I only have a small comment

jfaltermeier (approved):

PR #281, opened 2024-02-14
[terraform] Add configuration and cluster creation for kubeadm.

inlann:
Hello,

Glad to be able to contribute!
I added a set of Terraform configurations to allow users to deploy Theia-Cloud on their own Kubernetes clusters created through kubeadm.


jfaltermeier:
Thank you! I will be having a look in the next days and give it a try

jfaltermeier:
I've had a first look and I am struggling setting up a working cluster using kubeadm, but this is not related to your changes. 

As far as I can see you mainly modify the kubernetes, kubectl, and helm providers to use the default paths for the certificates/key files as used/created by kubeadm. 
The configuration does not create the cluster via terraform, which is a difference from the other getting started guides. 

The providers all allow to be configured by pointing to the local kube config file as well:
https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#file-config
https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs#load_config_file
https://registry.terraform.io/providers/hashicorp/helm/latest/docs#example-usage

I am wondering if it would be easier/more useful to create a getting started that allows to use any existing cluster just based on the user default kubectl configuration?
As a prerequisites users would have to make sure that their default config points to the desired cluster. This should also work for your use case then. 
Also this would skip having to add the `cluster_host` in `terraform/configurations/kubeadmin_getting_started/kubeadm_getting_started.tf` manually

inlann:
Thanks for your review!

> I've had a first look and I am struggling setting up a working cluster using kubeadm on my test machine, but this is not related to your changes.

I create the cluster by running `kubeadm init --pod-network-cidr=192.168.0.0/16` since I use the calico as the CNI.

Before that, I have a terraform project using proxmox provider to create virtual machines for kubernetes, and a shell script to initialize all environment needed by the kubernetes such as containerd, runc and cni-plugin.

And then, I install longhorn as CSI and ingress-nginx using helm.

> As far as I can see you mainly modify the kubernetes, kubectl, and helm providers to use the default paths for the certificates/key files as used/created by kubeadm.
> The configuration does not create the cluster via terraform, which is a difference from the other getting started guides.

You're right. The `modules/cluster_creation/kubeadm` is copied from `modules/cluster_creation/minikube`. I found that the module for creating clusters will eventually only provide a few variables to theia-cloud such as cluster certificates and key and the cluster host address. So I put them into the `modules/cluster_creation/kubeadm/outputs.tf` directly.

Actually, this PR would be more appropriately named "Add configuration for an existing Kubernetes cluster."

BTW, I am also attempting to automate the creation of Kubernetes virtual machines from scratch using Terraform and to initialize the Kubernetes clusters within them. If this part can be of help to theia-cloud, I would be happy to share my solution.

inlann:
> The providers all allow to be configured by pointing to the local kube config file as well: https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#file-config https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs#load_config_file https://registry.terraform.io/providers/hashicorp/helm/latest/docs#example-usage
> 
> I am wondering if it would be easier/more useful to create a getting started that allows to use any existing cluster just based on the user default kubectl configuration? As a prerequisites users would have to make sure that their default config points to the desired cluster. This should also work for your use case then. Also this would skip having to add the `cluster_host` in `terraform/configurations/kubeadmin_getting_started/kubeadm_getting_started.tf` manually

That's awesome! I will simplify the PR with that.

PR #293, opened 2024-04-11
Add options field to CRDs

sgraband:
This field can be used by adopters to pass arbitrary data to the operator. 
Field is added to all three CRDs for maximum configurability. 
Bumped version number for the CRDs and adjusted the conversion webhook.

Helm repo PR: https://github.com/eclipsesource/theia-cloud-helm/pull/55


jfaltermeier (approved):
LGTM! I've only briefly tested adding options to sessions which worked, so I think the other CRs should be fine as well.

PR #306, opened 2024-06-06
Rework CI workflow names

sgraband:
Prefix each name with the type (MAVEN, DOCKER, NPM). 
Use short names for each component.
Remove the long name for workflow calls, simply call them `run`. 
Merge jobs together to narrow down the amount of jobs being called.

Screenshot:
image


lucas-koehler (approved):
LGTM thanks! The namings are much cleaner now :sparkles:

PR #308, opened 2024-06-06
NPM workflow update and docker, maven workflow fix

lucas-koehler:
Improve the reusable npm workflow:
- Configure npm provenance for package publishing
- Update setup-node action to v4
- Relax node version to any Node 20 version. Our docker images behave the same.
- Explicitly set permissions for workflow
- Get package version from package.json instead of hardcoded env variable

Fix reusable docker and maven workflows: Add missing `fi`s


sgraband (approved):
Thanks! LGTM :+1:

PR #314, opened 2024-06-11
Rename Theia.Cloud to Theia Cloud and Improve Ingress/Host configuration

jfaltermeier:
* regenerate common API using older version of opanapi generator
* also initialize Session Status last reported activity with currentTimeMillis. There is a race condition that if this is checked shortly after creation, than it is seen as 1970 and session is stopped immediately. 
* adjust to helm chart changes wrt Ingress/Host configuration

Java Packages / CRDs / Maven configurations will be renamed once we have a final decision on future naming

https://github.com/eclipsesource/theia-cloud/pull/314
https://github.com/eclipsesource/theia-cloud-helm/pull/59
https://github.com/eclipsesource/theia-cloud-website/pull/33


sgraband (approved):
LGTM :+1:

Some points were we should not forget to refactor:
https://github.com/eclipsesource/theia-cloud/blob/5a5538b163c82702048af92e6ff360b9080c93c2/node/monitor-theia/src/common/monitor-protocol.ts#L7
https://github.com/eclipsesource/theia-cloud/blob/5a5538b163c82702048af92e6ff360b9080c93c2/node/monitor/src/modules/activity-tracker-module.ts#L10
https://github.com/eclipsesource/theia-cloud/blob/5a5538b163c82702048af92e6ff360b9080c93c2/node/common/src/client.ts#L187

PR #322, opened 2024-07-08
Add publishing to Google artifact registry

sgraband:
Pass secrets from parent workflows.
Add profile for publishing to Google.
Adjust workflow to publish to both Google and Github.


jfaltermeier (approved):
LGTM

PR #323, opened 2024-07-15
Move default behavior into library

sgraband:
Make code reuseable and reduce complexity of minimal example (defaultoperator). 
Make leader election optional by making it not the default implementation, but a optional class.

Note: This is just a proposal, please feel free to comment if you would like to structure this differently or have any other objections.


sgraband:
@lucas-koehler Thanks for the feedback. I incorporated both of your suggestions. Let me know what you think.

lucas-koehler (approved):
Thanks for the updates! LGTM now :sparkles:

sgraband:
@jfaltermeier Do you want to take a look as well?

java/operator/org.eclipse.theia.cloud.defaultoperator/src/main/java/org/eclipse/theia/cloud/defaultoperator/DefaultTheiaCloudOperatorLauncher.java
```diff
@@ -15,122 +15,18 @@
...
-import picocli.CommandLine;
-
-public class DefaultTheiaCloudOperatorLauncher implements TheiaCloudOperatorLauncher {
-
-    private static final Logger LOGGER = LogManager.getLogger(DefaultTheiaCloudOperatorLauncher.class);
-
-    protected static final String LEASE_LOCK_NAME = "theia-cloud-operator-leaders";
+public class DefaultTheiaCloudOperatorLauncher extends LeaderElectionTheiaCloudOperatorLauncher {
 
     static final String COR_ID_INIT = "init";
```
jfaltermeier, line 24: I think this constant can be removed?

PR #325, opened 2024-07-18
Standardize & apply standard formatting

### LINKED ISSUE #152: Standardize & apply standard formatting
In a lot of places in the code, there's a weird mix of tabs & spaces, causing function bodies not to be indented correctly.

For an example, see [LazySessionHandler.java](https://github.com/eclipsesource/theia-cloud/blob/main/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/impl/LazySessionHandler.java)

It seems that tab size is assumed to be 8 spaces, but then again in a lot places 4 spaces are used for indentation.

I propose adding a convention for the project's code, and perhaps a code formatter (perhaps as a Maven plugin?)

sgraband:
Standardize the formatting for the use in VSCode.
Use the `java-formatter.xml` file together with the `vscjava.vscode-java-pack` for Java formatting.
Define prettier as the default formatter for most of the files.
Exceptions include:
- dockerfiles are formatted using `ms-azuretools.vscode-docker`
- plantuml files are formatted using `jebbs.plantuml`

Also add all needed extensions to the recommendations.
Furthermore, transform all launch configs to tasks & debug configs.

Fixes #152


jfaltermeier:
I pushed an additional commit. Some of the xml files still had tabs. 
VS Code did not update these, the Theia IDE did. After it was fixed, VS Code also produces spaces instead of tabs. 
Not sure what was/is the issue there. 
Could you maybe shortly check if you can also change the xml files now without any additional changes?

jfaltermeier (approved):
In general I think it looks good. 
Formatting seems to work and the the debug configurations / tasks also work. 
If we notice any inconsistencies, we need to log new issues. 

In general we should probably also add more documentation on how to run the theia cloud services locally, because e.g. the conversion webhook currently fails with `java.nio.file.NoSuchFileException: /etc/webhook/certs/tls.crt`, since this path is hard-coded in the `application.properties`

But this is out of context for this PR

sgraband:
@jfaltermeier Thanks for the commit. For me both Theia IDE and VSCode do not change the formatting with tabs/whitespaces. So i can no longer reproduce that issue, so we should be good to go. More documentation regarding building locally definately makes sense, but i also agree that this should be done in a follow up :+1:

PR #326, opened 2024-07-19
Rework Kubernetes resource naming and set up common unit tests

### LINKED ISSUE #324: Improve Kubernetes resource naming
Currently, naming of session resources (deployment, pod, config maps, etc) are not very useful. They start with a UUID which takes up most of the name and is followed by some part of the user and app definition.

We should improve naming to consistently start with session and have fixed length segments for user, app definition and part of a uuid.

Something like: `session-<user>-<appdef>-<uuid>` with each segment being limited in characters to not exceed the maximum length of 63 characters.

lucas-koehler:
Fixes #324 

Note: This will need a rebase if the formatting PR #325  is merged first. Otherwise, the touched Java files need to be formatted in the PR, too.

Rework the naming of Kubernetes resources for workspaces and sessions in the common package used by operator and service.

The goal is to make resource names more readable and contain useful information. Resource names now contain a prefix, optional identifier, user, app definition and the last segment of the source CRs (session, workspace, app definition) UID.

Also configure unit tests using JUnit 5 for the common maven module and add some tests for the new name generation.

---

Notes:
- Most suffixes and prefixes containing the resource type were removed because this is already given by the type.
- The new name calculation limits names to length 57 to leave 6 characters for Kubernetes to append a hash to deployment pods. This is at least 5 characters plus a dash. If there is not enough space in the resource name, the end of the pod name is replaced with this. Thus, the limit of 57 characters to avoid this.
- Changed the identifiers for the email and proxy config maps to `email` resp. `proxy` to be as short and concise as possible.

---

See an example of created resources for a user `bar` with app definition `theia-cloud-demo`:
image

### DIFF
--- a/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
+++ b/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
@@ -146,8 +146,8 @@ public final class NamingUtil {
          * This must be shorter than {@link NamingUtil.VALID_NAME_LIMIT} We fill remaining space with additional
          * information about the workspace. This may be trimmed away however.
          */
-        return createName("workspace", identifier, workspace.getSpec().getUser(),
-                workspace.getSpec().getAppDefinition(), workspace.getMetadata().getUid());
+        return createName("ws", identifier, workspace.getSpec().getUser(), workspace.getSpec().getAppDefinition(),
+                workspace.getMetadata().getUid());
     }
 
     /**
--- a/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
+++ b/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (C) 2022 EclipseSource and others.
+ * Copyright (C) 2022-2024 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -15,7 +15,9 @@
  ********************************************************************************/
 package org.eclipse.theia.cloud.common.util;
 
+import java.util.Arrays;
 import java.util.Locale;
+import java.util.Objects;
 
 import org.eclipse.theia.cloud.common.k8s.resource.appdefinition.AppDefinition;
 import org.eclipse.theia.cloud.common.k8s.resource.session.Session;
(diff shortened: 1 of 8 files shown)


java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
```diff
@@ -129,45 +178,86 @@ public static String createName(Workspace workspace, String identifier) {
 	 * We fill remaining space with additional information about the workspace. This
 	 * may be trimmed away however.
 	 */
-	return createName(workspace.getMetadata().getUid(), identifier, getAdditionalInformation(workspace),
-		MAX_IDENTIFIER_LENGTH);
+	return createName("workspace", identifier, workspace.getSpec().getUser(),
```
sgraband, line 181:
```suggestion
	return createName("ws", identifier, workspace.getSpec().getUser(),
```

Just as an idea, would `ws` be enough? We use it as a shortcut for `workspace` so it should be clear what is meant, and there is more space, for the user and appDef name?

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
```diff
@@ -129,45 +178,86 @@ public static String createName(Workspace workspace, String identifier) {
...
+	// If the user is an email address, only take the part before the @ sign because
+	// this is usually sufficient to identify the user.
+	String userName = user.split("@")[0];
+
+	int infoSegmentLength;
+	String shortenedIdentifier = null;
+	if (identifier == null || identifier.isBlank()) {
+	    infoSegmentLength = 17;
+	} else {
+	    infoSegmentLength = 11;
```
sgraband, line 222: Where do these two values come from? I think this could use a comment. Also do we not need to take the length of the prefix into consideration?

lucas-koehler: The number came from calculating how long segments can be for at most 57 characters in total, with `session-` prefix. The prefix is shortened to a max length.
This stems from me implementing this first for sessions only and the reusing that for workspace and app definition.

I guess we could also calculate the segment length dymically based on the prefix length and potentially a max character limit (i.e. 57 or 63).

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/NamingUtil.java
lucas-koehler: You are right. I thought `workspace` is more readable. However someone accessing the resources with kubectl is probably knowledgable enough to understand the ws- prefix, too

lucas-koehler:
@sgraband I rebased the changes and shortened the workspace prefix to `ws`. I did not improve the algorithm to use the space fully, yet because I'm out of time. I suggest doing this in a follow up so we can merge this already.

sgraband (approved):
Thanks! LGTM :+1:

PR #328, opened 2024-07-19
Check Workspace Owner when launching non-ephemeral Sessions

jfaltermeier:
The Service does filter out workspaces if the user does not match via https://github.com/eclipsesource/theia-cloud/blob/37785cfeb9a8b4cebd79e404750b74c2d8ec30a1/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/K8sUtil.java#L127

This change just makes sure that if there was a bug in the service, the operator does not just expose the workspace belonging to a different user.


sgraband (approved):
Thanks LGTM :+1:

PR #331, opened 2024-07-23
Fix maven publishing for release versions

sgraband:
Add the repository field to the pom.xml.
Also try to fix that the build dependencies step is not executed on release version pushes.

_Note: I manually pushed the 0.11.0 versions of the maven artifacts to the Google Artifact Registry._


jfaltermeier (approved):
LGTM

PR #335, opened 2024-07-23
Increase client_timeout for terraform keycloak provider

jfaltermeier:
should avoid timeouts when connecting to keycloak


sgraband (approved):
LGTM :+1:

PR #345, opened 2024-08-21
Update Dependencies

jfaltermeier:
## Java

Chosen versions for k8s client and quarkus mainly based on the latest Webhook Framework: https://mvnrepository.com/artifact/io.javaoperatorsdk/kubernetes-webhooks-framework-core/1.1.4

-> uses kubernetes client 6.12.1
-> quarkus 3.11.3 uses 6.12.1

This is the test matrix for the client 6.12.1:
https://github.com/fabric8io/kubernetes-client/blob/e4ddf5596385374fa2506ecf255b185629bc0b2e/.github/workflows/e2e-tests.yml#L71

## Node

Updated open api generator and axios
-> common library should be tested before merging

Updated express
-> monitors should be tested

---


lucas-koehler (approved):
Code changes LGTM :partying_face: The testing page - using the common module - still worked well. Both monitor extensions also still work: They returned timestamps and showed the notification popup

PR #354, opened 2024-09-27
Fix lastActivity not being initialized properly

sgraband:
- Improve and document debugging of operator and service
- Minor improvement: Local testing values should only pull IfNotPresent
- Fix lastActivity not being initialized properly
    - We need to set it after the resource has been created.
    - Therefore it is now handled by the operator not the service.
    - Add warning in case resources are edited before being applied to the cluster.
    - Only check HANDLED sessions from the activity tracking.
    - Other resources should be cleaned up differently.

### DIFF
--- a/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/plugins/MonitorActivityTracker.java
+++ b/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/plugins/MonitorActivityTracker.java
@@ -29,10 +29,12 @@ import java.util.Optional;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.eclipse.theia.cloud.common.k8s.client.TheiaCloudClient;
+import org.eclipse.theia.cloud.common.k8s.resource.OperatorStatus;
 import org.eclipse.theia.cloud.common.k8s.resource.appdefinition.AppDefinition;
 import org.eclipse.theia.cloud.common.k8s.resource.session.Session;
 import org.eclipse.theia.cloud.operator.TheiaCloudOperatorArguments;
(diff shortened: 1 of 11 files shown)


sgraband:
@lucas-koehler Could you review this?

java/service/org.eclipse.theia.cloud.service/README.md
```diff
@@ -14,6 +14,14 @@ You can run your application in dev mode that enables live coding using:
 
 > **_NOTE:_** Quarkus now ships with a Dev UI, which is available in dev mode only at <http://localhost:8080/q/dev/>.
 
+## Debug the service
+
+When you have the service running in dev mode you can use the `Attach to Service` debug config to debug the service.
+
+Check the [OpenAPI documentation](../../../documentation/OpenAPI.md) to see how you can send requests to the service.
+
+_Note_: The landing page will not communicate with this service, as it will be hosted on another address.
```
lucas-koehler, line 23: Maybe it makes sense to point readers to the test page here? With this, various requests to the service could still be tested then :)

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/plugins/MonitorActivityTracker.java
```diff
@@ -78,7 +80,10 @@ public void start() {
     }
 
     protected void pingAllSessions() {
-        List<Session> sessions = resourceClient.sessions().list();
+        // Only look at handled sessions (handled sessions have a lastActivity)
+        List<Session> sessions = resourceClient.sessions().list().stream()
+                .filter(session -> OperatorStatus.HANDLED.equals(session.getStatus().getOperatorStatus()))
+                .collect(Collectors.toList());
```
lucas-koehler, line 86: Optional: We can simplify this leveraging the toList() method introduced with Java 16.
```suggestion
                .toList();
```

lucas-koehler (approved):
Hi @sgraband the code changes and the updated documentation look good to me :sparkles: I added two small non-blocking remarks inline

sgraband:
Thanks for the comments/suggestions @lucas-koehler. I added both of them. Could you take another quick look?

lucas-koehler (approved):
Thanks for the updates. LGTM :sparkle:

PR #362, opened 2024-10-22
Improve session pod labels

xai:
org.eclipse.theia.cloud.common.util.LabelsUtil is added to encapsulate creation of labels.

To indicate that a pod is a user session, its
'app.kubernetes.io/component' label is set to 'session'.

For custom labels, we use the prefix 'theia-cloud.io'.

This is currently used for:
- Identifying a session pods associated with a user: 'theia-cloud.io/user' = '$username'
- Identifying a session pod by appdefinition name: 'theia-cloud.io/app-definition' = $appdefinitionname

#366

### DIFF
--- a/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
+++ b/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
@@ -1,0 +1,26 @@
+package org.eclipse.theia.cloud.common.util;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.theia.cloud.common.k8s.resource.appdefinition.AppDefinitionSpec;
+import org.eclipse.theia.cloud.common.k8s.resource.session.SessionSpec;
+
+public class LabelsUtil {
+    public static final String LABEL_CUSTOM_PREFIX = "theia-cloud.io";
+
+    public static final String LABEL_KEY_SESSION = "app.kubernetes.io/component";
+    public static final String LABEL_VALUE_SESSION = "session";
+
+    public static final String LABEL_KEY_USER = LABEL_CUSTOM_PREFIX + "/user";
+    public static final String LABEL_KEY_APPDEF = LABEL_CUSTOM_PREFIX + "/app-definition";
+
+    public static Map<String, String> createSessionLabels(SessionSpec sessionSpec, AppDefinitionSpec appDefinitionSpec) {
+        Map<String, String> labels = new HashMap<>();
+        labels.put(LABEL_KEY_SESSION, LABEL_VALUE_SESSION);
+        String userName = sessionSpec.getUser().split("@")[0];
+        labels.put(LABEL_KEY_USER, userName);
+        labels.put(LABEL_KEY_APPDEF, appDefinitionSpec.getName());
+        return labels;
+    }
+}
--- a/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
+++ b/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
@@ -12,14 +12,19 @@ public class LabelsUtil {
     public static final String LABEL_KEY_SESSION = "app.kubernetes.io/component";
     public static final String LABEL_VALUE_SESSION = "session";
 
+    public static final String LABEL_KEY_THEIACLOUD = "app.kubernetes.io/part-of";
+    public static final String LABEL_VALUE_THEIACLOUD = "theia-cloud";
+
     public static final String LABEL_KEY_USER = LABEL_CUSTOM_PREFIX + "/user";
     public static final String LABEL_KEY_APPDEF = LABEL_CUSTOM_PREFIX + "/app-definition";
 
-    public static Map<String, String> createSessionLabels(SessionSpec sessionSpec, AppDefinitionSpec appDefinitionSpec) {
+    public static Map<String, String> createSessionLabels(SessionSpec sessionSpec,
+            AppDefinitionSpec appDefinitionSpec) {
         Map<String, String> labels = new HashMap<>();
         labels.put(LABEL_KEY_SESSION, LABEL_VALUE_SESSION);
-        String userName = sessionSpec.getUser().split("@")[0];
-        labels.put(LABEL_KEY_USER, userName);
+        labels.put(LABEL_KEY_THEIACLOUD, LABEL_VALUE_THEIACLOUD);
+        String sanitizedUser = sessionSpec.getUser().replaceAll("@", "_at_").replaceAll("[^a-zA-Z0-9]", "_");
+        labels.put(LABEL_KEY_USER, sanitizedUser);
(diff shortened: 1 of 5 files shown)


java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
```diff
@@ -0,0 +1,26 @@
...
+    public static final String LABEL_KEY_APPDEF = LABEL_CUSTOM_PREFIX + "/app-definition";
+
+    public static Map<String, String> createSessionLabels(SessionSpec sessionSpec, AppDefinitionSpec appDefinitionSpec) {
+        Map<String, String> labels = new HashMap<>();
+        labels.put(LABEL_KEY_SESSION, LABEL_VALUE_SESSION);
+        String userName = sessionSpec.getUser().split("@")[0];
+        labels.put(LABEL_KEY_USER, userName);
+        labels.put(LABEL_KEY_APPDEF, appDefinitionSpec.getName());
+        return labels;
+    }
```
lucas-koehler, line 25: While we already add the `app.kubernetes.io/component` annotation, I think we should also add the `app.kubernetes.io/part-of` annotation with value `theia-cloud`.
This is helpful if Theia Cloud is installed together with other services in the same namespace. When we eventually apply this to all of Theia Cloud's components, this provides a selector to get everything related to Theia Cloud.

What do you think?

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
```diff
@@ -0,0 +1,26 @@
...
+    public static final String LABEL_KEY_SESSION = "app.kubernetes.io/component";
+    public static final String LABEL_VALUE_SESSION = "session";
+
+    public static final String LABEL_KEY_USER = LABEL_CUSTOM_PREFIX + "/user";
+    public static final String LABEL_KEY_APPDEF = LABEL_CUSTOM_PREFIX + "/app-definition";
+
+    public static Map<String, String> createSessionLabels(SessionSpec sessionSpec, AppDefinitionSpec appDefinitionSpec) {
+        Map<String, String> labels = new HashMap<>();
+        labels.put(LABEL_KEY_SESSION, LABEL_VALUE_SESSION);
+        String userName = sessionSpec.getUser().split("@")[0];
```
lucas-koehler, line 21: Is there a reason we remove the second part of the email address (in case the user is an email address)? I would not do this because in deployments with a larger user base this can lead to collisions quite easily.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/EagerStartSessionHandler.java
```diff
@@ -115,6 +118,25 @@ public boolean sessionAdded(Session session, String correlationId) {
             return false;
         }
 
+        try {
+            client.services().inNamespace(client.namespace()).withName(serviceToUse.get().getMetadata().getName())
+                .edit(service -> {
+                    LOGGER.info("Setting pod labels");
```
lucas-koehler, line 124: I think this should only be a debug level log as nothing exiting or too relevant happens here functionality wise.

```suggestion
                    LOGGER.debug("Setting service labels");
```

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/LazySessionHandler.java
```diff
@@ -444,6 +446,16 @@ protected void createAndApplyDeployment(String correlationId, String sessionReso
         }
         K8sUtil.loadAndCreateDeploymentWithOwnerReference(client.kubernetes(), client.namespace(), correlationId,
                 deploymentYaml, Session.API, Session.KIND, sessionResourceName, sessionResourceUID, 0, deployment -> {
+
+                    LOGGER.info("Setting pod labels");
```
lucas-koehler, line 450: I think this should only be a debug level log as nothing exiting or too relevant happens here functionality wise.
```suggestion
                    LOGGER.debug("Setting pod labels");
```

lucas-koehler (changes requested):
Hi @xai ,
Thanks for this improvement :) Generally the changes look pretty good to me already. I have  a few small remarks inline. Please have a look :)

lucas-koehler:
> In the long run, I think we should consider adding the same labels to all deployments, services, and configmaps that are created. 

I agree: Eventually we should label _all_ components created by Theia Cloud, either by the operator or the Helm charts.

> I think we may also ignore the EagerStartSessionHandler for now.
I am not sure if we should remove this handler for 1.0.0, since it does not cover all functionality anyway.

I think we should leave it in but make very clear in the documentation on the website, the helm charts and potentially here that the implementation is not up to par. My reason for leaving it in is that it might encourage contributions to improve this and shows our intent on bringing this up to par eventually. Furthermore, it allows basic prototyping of eager start. WDYT?

jfaltermeier:
> I think we should leave it in but make very clear in the documentation on the website, the helm charts and potentially here that the implementation is not up to par. My reason for leaving it in is that it might encourage contributions to improve this and shows our intent on bringing this up to par eventually. Furthermore, it allows basic prototyping of eager start. WDYT?

Sounds good

java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
xai: I adjusted that to use the full email address. I had to translate the `@` because labels must not contain that character.

jfaltermeier (approved):
LGTM

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/util/K8sUtil.java
```diff
@@ -164,6 +170,21 @@ private static <T extends HasMetadata, U, V extends Resource<T>> Optional<T> loa
...
 
+            // Apply labels to the resource metadata
+            if (newItem.getMetadata().getLabels() == null) {
+                newItem.getMetadata().setLabels(new HashMap<>());
+            }
+            newItem.getMetadata().getLabels().putAll(labelsToAdd);
+
+            // If the resource is a Deployment, also apply labels to the pod template metadata
+            if (newItem instanceof Deployment) {
+                Deployment deployment = (Deployment) newItem;
```
lucas-koehler, line 181: Minor: As we use Java 17, you should be able to simplify this:
```suggestion
            if (newItem instanceof Deployment deployment) {
```

lucas-koehler (approved):
Code changes LGTM me. I have one very minor, non-blocking suggestion inline.
I did not retest the second iteration though. @jfaltermeier Did you test it? Otherwise, I can still do it.

jfaltermeier:
> Code changes LGTM me. I have one very minor, non-blocking suggestion inline. I did not retest the second iteration though. @jfaltermeier Did you test it? Otherwise, I can still do it.

I did a short test and saw the labels on pods, deployments, services, and config maps

PR #363, opened 2024-10-24
Add REST API markdown docs based on openapi definition

lucas-koehler:
Use the openapi-generator-cli to generate markdown docs from the openapi.json.
Also document how this works and add a hint that the generator supports
various doc and code languages.


jfaltermeier (approved):
LGTM. Regeneration worked on my side as well.

The generated markdown files don’t follow our formatting rules, but since we shouldn’t be editing generated files manually anyway, I’m not sure if it’s an issue. What’s your take on this?

lucas-koehler:
> The generated markdown files don’t follow our formatting rules, but since we shouldn’t be editing generated files manually anyway, I’m not sure if it’s an issue. What’s your take on this?

We could format them using prettier on the command line. However, I don't see any reason to do this for now because - as you stated - these files should not be edited manually anyway. Thus, I would not add a step in the documentation for now.

PR #371, opened 2024-11-04
Update dependencies

jfaltermeier:
* rollup
* vite
* body-parser
* express
* path-to-regexp
* send
* serve-static
* micromatch
* webpack
* cookie
* junit in demos
* update workflow actions (use sha for non github actions)
* update helm dependencies to latest version (but keycloak)

### DIFF
--- a/node/landing-page/package.json
+++ b/node/landing-page/package.json
@@ -1,7 +1,7 @@
 {
   "name": "landing-page",
   "version": "0.1.0",
-  "license": "EPL-2.0",
+  "license": "EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0",
   "private": true,
   "type": "module",
   "scripts": {
--- a/node/landing-page/package.json
+++ b/node/landing-page/package.json
@@ -1,6 +1,7 @@
 {
   "name": "landing-page",
   "version": "0.1.0",
+  "license": "EPL-2.0",
   "private": true,
   "type": "module",
   "scripts": {
--- a/node/testing-page/package.json
+++ b/node/testing-page/package.json
@@ -1,6 +1,7 @@
 {
   "name": "testing-page",
   "version": "0.1.0",
+  "license": "EPL-2.0",
   "private": true,
   "dependencies": {
     "@eclipse-theiacloud/common": "1.0.0-next",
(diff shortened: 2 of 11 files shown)


node/landing-page/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "landing-page",
   "version": "0.1.0",
+  "license": "EPL-2.0",
```
lucas-koehler, line 4: I think our license actually is the following. We also use this in headers in the operator java files and the LICENSE file. WDYT?
```suggestion
  "license": "EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0",
```

node/testing-page/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "testing-page",
   "version": "0.1.0",
+  "license": "EPL-2.0",
```
lucas-koehler, line 4: I think our license actually is the following. We also use this in headers in the operator java files and the LICENSE file. WDYT?
```suggestion
  "license": "EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0",
```

lucas-koehler (commented):
The changes already look quite good to me :)
I have 2 small remarks inline regarding the license.

I wonder whether we should also update the maven dependencies in https://github.com/eclipse-theia/theia-cloud/blob/main/java/common/maven-conf/pom.xml
I.e. quarkus, maven surefire, etc. Was this missed or left out on purpose?

Note that I did not test this as the changes are pretty small. Let me know if I should.

jfaltermeier:
> The changes already look quite good to me :) I have 2 small remarks inline regarding the license.
> 
> I wonder whether we should also update the maven dependencies in https://github.com/eclipse-theia/theia-cloud/blob/main/java/common/maven-conf/pom.xml I.e. quarkus, maven surefire, etc. Was this missed or left out on purpose?
> 
> Note that I did not test this as the changes are pretty small. Let me know if I should.

I checked for a new version of https://mvnrepository.com/artifact/io.javaoperatorsdk/kubernetes-webhooks-framework-core at the beginning of November. I usually pick the corresponding io.fabric8 k8s-client version from this and update Quarkus accordingly, ensuring all dependencies are built with the same or a similar k8s-client version.

Since there wasn't a release at this time, I mainly focused on Dependabot alerts for Maven.

There’s a new version available now, so I’ll update to it after 1.0.

lucas-koehler (approved):
Thanks for the update and explanation. LGTM :sparkles:

PR #393, opened 2025-01-10
Fix and Improve Eager Start Handling

lucas-koehler:
Requires:
- https://github.com/eclipse-theia/theia-cloud-helm/pull/86

## Summary
This pull request addresses critical issues in the eager start handling, which was previously broken, and introduces improvements to naming, session management, pod synchronization, and performance in Kubernetes environments.

### Common
- Name Handling for Eager Start:
  - Resolved an endless loop and a NullPointerException (NPE) in NamingUtil methods.
  - Fixed ID extraction from generated names to align with updates from an [earlier PR](https://github.com/eclipse-theia/theia-cloud/pull/326).

### Operator - Eager Session Handling

- Added annotations to trigger ConfigMap updates for OAuth proxy synchronization in pods. This ensures, the assigned user can actually access the pod immediately after redirection
- Implemented session cleanup and pod restart logic in the renamed EagerSessionHandler.

### Service

- Enhanced max instance checks to consider sessions relevant to the app definition instead of analyzing all pods.
- Updated pod resolution logic to rely on owner references instead of environment variables, ensuring compatibility with eager start mode.
- Improved logging and performance reporting in K8sUtil and the performance REST handler.

### Testing

- Added a Terraform test configuration for eager start mode.
- Added a VS Code launch config to start the operator in eager mode for the new Terraform config
- Extend testing page with a button to get session performance data

## Notes on Testing

- You need the corresponding Theia Cloud Helm changes from PR https://github.com/eclipse-theia/theia-cloud-helm/pull/86 for the changes to work as they require additional K8S permissions for the operator and service.
- I think all of our test setups currently do not configure snippet enablement for the nginx ingress of Minikube. This must be done manually. See here how to do this: https://github.com/eclipse-theia/theia-cloud-helm?tab=readme-ov-file#cluster-prerequisites . We should probably fix this in general in a separate PR.
- The default Theia Cloud IDE app is not supposed to work in eager mode as it has 0 `minInstances` and currently lazy starting up until maxInstances is not supported (and maybe never will be). Use the CDT cloud app.

### DIFF
--- a/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/EagerSessionHandler.java
+++ b/java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/EagerSessionHandler.java
@@ -30,6 +30,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.eclipse.theia.cloud.common.k8s.client.TheiaCloudClient;
 import org.eclipse.theia.cloud.common.k8s.resource.appdefinition.AppDefinition;
+import org.eclipse.theia.cloud.common.k8s.resource.appdefinition.AppDefinitionSpec;
 import org.eclipse.theia.cloud.common.k8s.resource.session.Session;
 import org.eclipse.theia.cloud.common.k8s.resource.session.SessionSpec;
 import org.eclipse.theia.cloud.common.util.JavaUtil;
@@ -45,7 +46,9 @@ import org.eclipse.theia.cloud.operator.util.TheiaCloudServiceUtil;
 
 import com.google.inject.Inject;
 
+import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.Service;
+import io.fabric8.kubernetes.api.model.ServiceList;
 import io.fabric8.kubernetes.api.model.networking.v1.HTTPIngressPath;
 import io.fabric8.kubernetes.api.model.networking.v1.HTTPIngressRuleValue;
 import io.fabric8.kubernetes.api.model.networking.v1.Ingress;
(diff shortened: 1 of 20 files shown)


java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/EagerSessionHandler.java
```diff
@@ -266,4 +306,140 @@ protected Ingress addIngressRule(Ingress ingress, Service serviceToUse, String h
         return ingress;
     }
 
+    @Override
+    public boolean sessionDeleted(Session session, String correlationId) {
```
jfaltermeier, line 310: On session delete, I noticed that the Ingress does not get patched. As a result, the session stays accessible until another user starts a new session and gets this one.

jfaltermeier (changes requested):
Looks very nice so far. Starting sessions works again. During testing I noticed an issue with deletion

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/EagerSessionHandler.java
lucas-koehler: Good catch, the reason is that resolving the service via the dangling owner reference fails:
While multiple researches via ChatGPT state that owner references are not automatically cleaned up and resolving the Service like this has worked for me before, it does not now anymore.

I can also observe that the owner references from the service and deployment to the deleted session are no longer shown using kubectl. Thus, they seem in fact to be cleaned up automatically - at least in Minikube.

I think it could be a timing issue whether the operator code runs before this happened. This would explain how it ran into code after this code:

```
Optional<Service> ownedService = TheiaCloudServiceUtil.getServiceOwnedBySession(sessionResourceName,
                sessionResourceUID, services);

        if (ownedService.isEmpty()) {
            LOGGER.error(
                    formatLogMessage(correlationId, "No Service owned by session " + sessionResourceName + " found."));
            return false;
        }
```

Nevertheless, the `sessionDeleted` method must be stable to work anyway. I'll investigate how to do this independent of the owner references. I guess an optional deletion of the references (if present) could make sense. I could imagine various versions and/or implementations of the kube-controller-manager behave differently in this regard.

lucas-koehler:
Hi @jfaltermeier , I added one commit that should fix session delete handling even if dangling owner references are cleaned up automatically. Please have a look again :)

PR #394, opened 2025-01-16
Fix reusable-theia-extension workflow

lucas-koehler:
Remove hardcoded `extensions/` prefix to align with the other commands using workflow input `package_workspace`


jfaltermeier (approved):
LGTM

PR #396, opened 2025-01-24
Stabilize service cleanup for eager start

lucas-koehler:
Removing the labels and owner reference fails for rare cases when cleaning an eager instance's service after a session was removed. This adds retries to stabilize the sessionDeleted handling.


jfaltermeier (approved):
Thanks. code looks good to me, but I haven’t tested.

PR #400, opened 2025-02-11
service: Add admin user concept and app definition scaling endpoint

lucas-koehler:
# Summary

This PR adds the concept of admin users to the Theia Cloud REST service. They are identified by a configurable admin group name in MicroProfile JWT's groups claim.
Admin only endpoints and resources can be configured via new annotation `@AdminOnly`. This is used for a new endpoint that allows updating an app definition's min and max instances. In addition an admin ping endpoint was added.
The terraform Keycloak module was extended to configure the default admin group and export realm, group, and test user infos for further usage in consuming modules.

A corresponding PR for the Helm chart allows configuring the admin group name as a value: https://github.com/eclipse-theia/theia-cloud-helm/pull/87

# How to test

- Setup Minikube using the terraform test configurations. To use a configure a custom group name for the service set value `keycloak.adminGroup` on the theia-cloud Helm chart.
  - You'll need to rename the group created by the Keycloak module manually in Keycloak's admin UI. You can find it in the TheiaCloud realm under Groups.
- You'll need an access token for the user `foo` which is configured to be an admin user by the terraform test setup
```bash
curl -s --insecure --request POST \
  --url https://$(minikube ip).nip.io/keycloak/realms/TheiaCloud/protocol/openid-connect/token \
  --header 'content-type: application/x-www-form-urlencoded' \
  --data grant_type=password \
  --data client_id=theia-cloud \
  --data username=foo \
  --data password=foo | jq .access_token
```
- Use this as a Bearer token in your request
- A GET request to `https://$(minikube ip).nip.io/service/service/admin/asdfghjkl` should return true with a token for user foo. It should return `403` for user bar because bar is not an admin user
- Update an app definition at endpoint PATCH `$(minikube ip).nip.io/service/service/admin/appdefinition/<appdef-name>` . The payload is a JSON object like so:
```json
{
  "appId": "asdfghjkl",
  "minInstances": 2,
  "maxInstances": 4
}
```

# Implementation details

<details><summary>Implementation details</summary>

## Add admin user support via annotation 

- Introduce the `@AdminOnly` annotation to mark REST resources or methods as accessible only to admin users.
- Implement AdminOnlyFilter that intercepts requests and aborts non-admin users with a 403 Forbidden response.
- Update ApplicationProperties to include a configurable admin group name property ("theia.cloud.auth.admin.group")
  with a default value of "theia-cloud/admin".
- Enhance TheiaCloudUser by adding an admin flag.
- Modify TheiaCloudUserProducer to derive the admin status from the MicroProfile JWT's groups claim.
- Add tests for the new admin-only filter, properties, and user producer functionality.
- Extend service Dockerfile to allow configuring the admin group name via environment variable.

## Add admin endpoint to update app def's min/max instances

- Add new resource AdminAppDefinitionAdminResource for all admin endpoints regarding app definitions
- Minor extensions in K8SUtil, AppDefinitionSpec to allow editing min/max instances
- Add tests for AdminAppDefintionAdminResource
- Add RootAdminResource with a ping endpoint that only returns if the user is an admin

## terrafom: Add admin user group and outputs for Keycloak module 

Extend Keycloak terraform module:
- Define an admin group with the default name
- Export realm, admin group and test users via outputs

Test setup:
- Add test user foo to the admin group

## fix: Update Keycloak URL in tasks.json to include realm path 

This is required for the Quarkus dev server to discover the OIDC config.
Also see the official documentation: <https://quarkus.io/guides/security-oidc-configuration-properties-reference#quarkus-oidc_quarkus-oidc-auth-server-url>
</details>

### DIFF
--- a/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/AdminOnlyFilter.java
+++ b/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/AdminOnlyFilter.java
@@ -42,8 +42,8 @@ public class AdminOnlyFilter implements ContainerRequestFilter {
     @Override
     public void filter(ContainerRequestContext requestContext) {
         if (!theiaCloudUser.isAdmin()) {
-            logger.infov("Blocked access to {0} {1} for non-admin user.", requestContext.getMethod(),
-                    requestContext.getUriInfo().getPath());
+            logger.infov("Blocked access to {0} {1} for non-admin user: {2}", requestContext.getMethod(),
+                    requestContext.getUriInfo().getPath(), theiaCloudUser.getIdentifier());
             requestContext.abortWith(Response.status(Response.Status.FORBIDDEN)
                     .entity("Admin privileges required to access this resource.").build());
         }
--- a/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/AdminOnlyFilter.java
+++ b/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/AdminOnlyFilter.java
@@ -1,0 +1,51 @@
+/********************************************************************************
+ * Copyright (C) 2025 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+package org.eclipse.theia.cloud.service;
+
+import org.jboss.logging.Logger;
+
+import jakarta.annotation.Priority;
+import jakarta.inject.Inject;
+import jakarta.ws.rs.Priorities;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerRequestFilter;
+import jakarta.ws.rs.ext.Provider;
+import jakarta.ws.rs.core.Response;
+
+/**
+ * ContainerRequestFilter aborting all requests from non admin users to resources annotated with {@link AdminOnly}.
+ */
+@Provider
+@AdminOnly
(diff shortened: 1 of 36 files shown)


java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/AdminOnlyFilter.java
```diff
@@ -0,0 +1,51 @@
...
+    @Inject
+    Logger logger;
+
+    @Inject
+    TheiaCloudUser theiaCloudUser;
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) {
+        if (!theiaCloudUser.isAdmin()) {
+            logger.infov("Blocked access to {0} {1} for non-admin user.", requestContext.getMethod(),
```
jfaltermeier, line 45: Should we also include the username/ID here? That way, we can see which user is trying to access the admin endpoint.

java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/admin/appdefinition/AppDefinitionUpdateRequest.java
```diff
@@ -0,0 +1,22 @@
...
+    @Schema(description = "The maximum number of instances to run.", required = false)
+    public Integer maxInstances;
+
+    public AppDefinitionUpdateRequest() {
+        super(KIND);
+    }
+
+    public AppDefinitionUpdateRequest(String appId) {
+        super(KIND, appId);
+    }
```
jfaltermeier, line 21: I think we should override `toString` to include more details about the update request.

Right now, the logs look like this:
`[4a887bae-4fe2-4e05-b94d-546bbb0b5e24] Update app definition ServiceRequest [appId=asdfghjkl, kind=appDefinitionUpdateRequest]`

jfaltermeier (commented):
Thanks, looks great! I just have two comments about the log messages.

Also, could you include this part of the testing instructions in a README?

```
curl -s --insecure --request POST \
  --url https://$(minikube ip).nip.io/keycloak/realms/TheiaCloud/protocol/openid-connect/token \
  --header 'content-type: application/x-www-form-urlencoded' \
  --data grant_type=password \
  --data client_id=theia-cloud \
  --data username=foo \
  --data password=foo | jq .access_token
```

I think it would be useful to have this easily accessible.

Besides that, we need to regenerate the openapi.json and probably also regenerate the common npm library + markdown documentation.

lucas-koehler:
Hi @jfaltermeier , thanks for the review. I addressed all issues mentioned in your review :)
I did not yet add the admin endpoints to the non-generated part of the common npm library due to time constraints and because the endpoints are not that useful without the corresponding operator implementations. IMO, we can expose them later there. The generated code already inlcudes them though.

jfaltermeier (approved):

PR #414, opened 2025-03-24
Backmerge 1.0.1 release / Changelog Update

jfaltermeier:
Make 1.0.1 visible on/reachable from main
From a changes perspective, we only want to merge the changelog update and keep all other files on main unchanged.


lucas-koehler (approved):
LGTM, thanks

PR #417, opened 2025-03-26
Include uid in session label

xai:
Since labels are potentially cropped, including the uid in the session label provides a reliable substring.

### DIFF
--- a/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
+++ b/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
@@ -1,3 +1,18 @@
+/********************************************************************************
+ * Copyright (C) 2024 EclipseSource.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
 package org.eclipse.theia.cloud.common.util;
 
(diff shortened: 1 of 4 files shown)


java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
```diff
@@ -33,15 +34,18 @@ private static String truncateLabelValue(String value) {
         return value;
     }
 
-    public static Map<String, String> createSessionLabels(SessionSpec sessionSpec,
-            AppDefinitionSpec appDefinitionSpec) {
+    public static Map<String, String> createSessionLabels(Session session, AppDefinitionSpec appDefinitionSpec) {
```
jfaltermeier, line 37: Thanks, looks good overall. I think we should pass `AppDefinition` instead of `AppDefinitionSpec` for consistency, even if we don't need it right now.
Also, could you please add this as a breaking change under the changelog for 1.1.0?

Regarding the labels, should we keep

`labels.put(LABEL_KEY_SESSION_NAME, truncateLabelValue(sessionSpec.getName()));`
and add an additional one like
`labels.put(LABEL_KEY_SESSION_UUID, truncateLabelValue(session.getMetadata().getUid()));`?

That way we'd have one label for unique identification and one that's more human-readable, but maybe truncated.

xai:
Added the missing license header in LabelsUtil.

jfaltermeier:
> ```
>  INFO: Running dash-licenses...
> usage: org.eclipse.dash.licenses.cli.Main [-batch <int>] [-confidence <int>]
>        [-excludeSources <sources>] [-help] [-project <shortname>] [-repo <url>]
>        [-review] [-summary <file>] [-timeout <seconds>] [-token <token>]
> ERROR: Command [
>   "java",
>   "-jar",
>   "/home/runner/work/theia-cloud/theia-cloud/node/node_modules/@eclipse-dash/nodejs-wrapper/download/dash-licenses.jar",
>   "package-lock.json",
>   "-batch",
>   50,
>   "-timeout",
>   200,
>   "-summary",
>   "/home/runner/work/theia-cloud/theia-cloud/node/license-check-summary.txt",
>   "-project",
>   "ecd.theia",
>   "-review"
> ] exit code (127) means dash-licenses has encountered an internal error
> ERROR: Detected an internal error in dash-licenses - run inconclusive
> Error: Process completed with exit code 127.
> ```
> 
> Any idea what went wrong here?

The Eclipse tool is unfortunately a bit unstable. However, since none of the files analyzed by the tool are touched and no new dependencies are introduced, I think we can safely ignore this error.

PR #418, opened 2025-03-27
Update NginX

jfaltermeier:
There were some CVEs reported for the nginx ingress controller.
We should update our default configuration to use a version that includes the relevant fixes


lucas-koehler (approved):
LGTM, thanks!

PR #439, opened 2025-07-29
Update form-data

jfaltermeier:
* 2.x.x version is not updated, because it is only a transient dependency via a test

### DIFF
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -10,7 +10,6 @@
       "workspaces": [
         "common",
         "e2e-tests",
-        "config-store",
         "landing-page",
         "testing-page"
       ],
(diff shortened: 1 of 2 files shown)


jfaltermeier:
This should fix some critical dependabot alerts

node/package-lock.json
```diff
@@ -10,8 +10,8 @@
       "workspaces": [
         "common",
         "e2e-tests",
+        "config-store",
```
lucas-koehler, line 13: This should not be here. I checked the node/package.json and it still has an entry for config-store in its workspaces. should we use this opportunity to rectify this ?

lucas-koehler (commented):
Hi @jfaltermeier , thanks for this update. Generally, the changes LGTM. I have one comment inline to rectify the node/package.json workspaces while we update the package-lock anyway.

lucas-koehler (approved):
Thanks for the update, LGTM now :)

PR #448, opened 2025-08-08
WIP - Update versions to 1.2.0-next/SNAPSHOT and Update Dependencies

jfaltermeier:


jfaltermeier:
Some hints:

Newer nginx-ingress controllers need this config to work with our snippet annotation:

  allow-snippet-annotations: 'true'
  annotations-risk-level: 'Critical'
  

Bitnami changes their docker images as well, we have to use 

  set {
    name  = "postgresql.image.repository"
    value = "bitnamilegacy/postgresql"
  }
  set {
    name  = "postgresql.volumePermissions.image.repository"
    value = "bitnamilegacy/os-shell"
  }
  
 values to get the old keycloak version. We want to update this anyway, but we also need to adapt config for this, so in the meantime we may use these legacy images

PR #467, opened 2025-11-21
Update node dependencies

jfaltermeier:
Focus was on removing the dependabot alerts for the demos and some more recent reports. 
Turns out that since https://github.com/eclipse-theia/theia-cloud/pull/432 we do not need the yarn.lock for the theia monitor demo anymore, so this will remove the reported alerts and we do not have to update this anymore. 

Also removes scheduled runs for the license check workflows, since they often fails because of rate limits. 

e2e tests run: https://github.com/eclipse-theia/theia-cloud/actions/runs/19566011783


lucas-koehler (approved):
LGTM, thanks for the updates and cleanup :)

PR #473, opened 2026-01-23
Support HAProxy ingress and configurable ingress path suffix

jfaltermeier:
* Add operator argument `--ingressPathSuffix` and apply it when creating/removing ingress rules
* Add Terraform `ingress_controller_type` switch (nginx default, haproxy optional) and wire it through configurations/modules
* Add HAProxy ingress controller Helm release and conditional Keycloak ingress annotations by ingress class
* Refactor Minikube test configs into staged steps (0_minikube-setup, 1_dependencies, 2_theia-cloud-base, 3_* apps) and update docs/outputs accordingly

### DIFF
--- a/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
+++ b/java/common/org.eclipse.theia.cloud.common/src/main/java/org/eclipse/theia/cloud/common/util/LabelsUtil.java
@@ -42,10 +42,14 @@ public class LabelsUtil {
 
     private static final int MAX_LABEL_LENGTH = 63;
 
-    private static String truncateLabelValue(String value) {
+    static String truncateLabelValue(String value) {
         if (value.length() > MAX_LABEL_LENGTH) {
             LOGGER.warning("Label value truncated: " + value);
-            return value.substring(0, MAX_LABEL_LENGTH);
+            String truncated = value.substring(0, MAX_LABEL_LENGTH);
+            while (truncated.length() > 0 && !Character.isLetterOrDigit(truncated.charAt(truncated.length() - 1))) {
+                truncated = truncated.substring(0, truncated.length() - 1);
+            }
+            return truncated;
         }
         return value;
     }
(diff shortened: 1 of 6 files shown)


terraform/test-configurations/1_dependencies/dependencies.tf
```diff
@@ -0,0 +1,152 @@
...
+}
+
+provider "kubernetes" {
+  host                   = data.terraform_remote_state.minikube.outputs.host
+  client_certificate     = data.terraform_remote_state.minikube.outputs.client_certificate
+  client_key             = data.terraform_remote_state.minikube.outputs.client_key
+  cluster_ca_certificate = data.terraform_remote_state.minikube.outputs.cluster_ca_certificate
+}
+
+resource "kubernetes_persistent_volume" "minikube" {
```
lucas-koehler: I think this should use `kubernetes_persistent_volume_v1` instead.
The current definition gives warnings like these:

```
 Warning: Deprecated Resource
│ 
│   with kubernetes_persistent_volume.minikube,
│   on dependencies.tf line 27, in resource "kubernetes_persistent_volume" "minikube":
│   27: resource "kubernetes_persistent_volume" "minikube" {
│ 
│ Deprecated; use kubernetes_persistent_volume_v1.
```

jfaltermeier:
> @lucas-koehler @jfaltermeier These changes break existing workspaces when the workspace name is too long (>54 char).
> 
> Consider we have a launch request with a long workspace name, e.g. `834c712b-80d8-411a-833d-e33f37b7827f_00db778f-9c0e-45df-8353-4ea96ca441d6` which already existed in the cluster. These long names worked before with the caveat that the workspace name in k8s was truncated if bigger than 62 char.
> 
> In [here](https://github.com/eclipse-theia/theia-cloud/blob/main/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/RootResource.java#L82) the service truncates the name with `asValidName(name, 62)` and uses it to check if the Workspace resource already exists in the cluster (this part was not changed, so it returns an existing workspace).
> 
> However when [creating the session](https://github.com/eclipse-theia/theia-cloud/blob/main/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/RootResource.java#L94), it uses the [UserWorkspace constructor](https://github.com/eclipse-theia/theia-cloud/blob/main/java/service/org.eclipse.theia.cloud.service/src/main/java/org/eclipse/theia/cloud/service/workspace/UserWorkspace.java#L58) which truncates the name with `asValidName(name, 62-8)`.
> 
> This results in the Session resource being created with spec.workspace truncated to 54 char.
> 
> The operator then fails to find the workspace resource and does not inject any volumes in the deployment.

Thanks for reporting this, I have opened [https://github.com/eclipse-theia/theia-cloud/issues/478](https://github.com/eclipse-theia/theia-cloud/issues/478).

PR #491, opened 2026-05-29
Update to Keycloak 26 and split Terraform modules

### LINKED ISSUE #472: Authentification failed on landing page with keycloak 26
### Describe the bug

My installation works fine with keycloak 24.0.3
But with 26.3.3 I'm getting Authentication Failed in assets/index-B0AV0o5v.js on landing page
Can theia cloud work with this version of keycloak? Is this issue related only to landing page or will sessions behave the same way?

### Expected behavior

Landing page working correctly

### Cluster provider

_No response_

### Version

_No response_

### Additional information

_No response_

lucas-koehler:
**Merge note**: This should probably be squashed on merge because there are various tryout and rebase-fix commits.

Fixes #472

E2E test run: https://github.com/eclipse-theia/theia-cloud/actions/runs/26575253687

# Update Keycloak setup and split Terraform modules

This PR updates the Keycloak deployment to Keycloak 26 (via the official Keycloak Operator instead of the Bitnami chart) and restructures the Terraform modules so that cluster prerequisites and Theia Cloud installation are clearly separated.

## Terraform: new `cluster-prerequisites` module

A new module replaces the prerequisite parts of the old `helm` module. It uses the official Keycloak Operator (instead of the deprecated Bitnami Helm chart) and is configurable enough to skip parts that are already installed in the target cluster.

- Installs the Keycloak Operator from `keycloak-k8s-resources` (default version `26.4.5`) and creates the Keycloak instance via the `Keycloak` CR
- Deploys an integrated PostgreSQL 17 backend (configurable, can be disabled to use an external database)
- Optionally installs cert-manager (`v1.17.4`) and a self-signed `ClusterIssuer` for local development
- Optionally installs the nginx or haproxy ingress controller with a configurable load balancer IP
- Creates the Keycloak `Ingress` with TLS, configurable cluster issuer, common name, annotations, and HTTP relative path (default `/keycloak/`)
- Exposes `keycloak_url`, `namespace`, `tls_secret_name`, and service-name outputs; `keycloak_url` is trimmed of any trailing slash so downstream consumers (e.g. the Keycloak provider) can use it directly
- Includes a README documenting variables, examples (Minikube, GKE, external Postgres, existing cert-manager) and a migration guide from the Bitnami chart

Two notable implementation details:

- The Keycloak Operator manifests are applied with `terraform_data` + `local-exec` instead of `kubectl_manifest`. The previous approach parsed YAML from an HTTP response to build dynamic `for_each` keys, which Terraform cannot evaluate at plan time.
- The Postgres PVC doesn't wait for a binding pod, because the pod that binds it is only created in a later step.

## Terraform: new `theia-cloud` module

A new module that only installs the Theia Cloud Helm charts (`theia-cloud-base`, `theia-cloud-crds`, `theia-cloud`). Each chart can be toggled individually via `install_theia_cloud_base`, `install_theia_cloud_crds`, and `install_theia_cloud` so that downstream configurations can stage installs as needed.

- New `theia_cloud_version` variable (default `1.2.0`) pins the chart version for all three releases
- New `keycloak_url` variable lets consumers pass the URL from the cluster-prerequisites module; falls back to `https://<hostname>/keycloak` if empty
- A single trailing slash is enforced on the URL before passing it to the Helm chart, since the chart requires it

## Terraform: removed `helm` module

The old `helm` module has been deleted; its functionality is now split between `cluster-prerequisites` and `theia-cloud`. 

## Terraform: updated configurations

All consumers have been migrated to the new modules:

- **`gke_getting_started`**: drops the unused `postgres_postgres_password` variable, installs cert-manager and the `theia-cloud-base` module first (to re-use its `letsencrypt-prod` `ClusterIssuer` for the Keycloak ingress), then `cluster-prerequisites`, then the remainder of `theia-cloud`. The Keycloak provider URL is taken from `module.cluster_prerequisites.keycloak_url`.
- **`minikube_getting_started/1_theiacloud-and-dependencies`**: replaces the `helm` module with `cluster_prerequisites` + `theia-cloud`. The standalone `.terraform.lock.hcl` at the configuration root has been removed (the configuration is split into sub-directories with their own lockfiles).
- **`ci-configurations/e2e_tests`**: switches to `cluster_prerequisites` for Keycloak and prerequisites; Theia Cloud charts continue to be installed from the local checkout via direct `helm_release`s.
- **`test-configurations/1_dependencies`**: switches to `cluster_prerequisites`. The `keycloak` output now exposes the module URL directly.

## Terraform: `keycloak` module

- Removes the now-redundant `hostname` variable (the Keycloak URL is supplied via the provider configuration)
- Sets `first_name`/`last_name` on the `foo` and `bar` test users so that first login does not prompt for them with Keycloak 26.

## Landing page and testing page (`keycloak-js` update)

Both pages have been updated for Keycloak 26 compatibility:

- `node/landing-page`: `keycloak-js` bumped from `20.0.5` to `26.2.4`
- `node/testing-page`: `keycloak-js` bumped from `^17.0.1` to `26.2.4`
- `App.tsx`: uses `new Keycloak(...)` (the v26 API no longer supports calling `Keycloak()` without `new`) and logs the underlying error in the auth failure handlers

## Documentation

- Top-level `README.md` release checklist now points to `terraform/modules/theia-cloud/variables.tf` for the Helm chart version (the old `terraform/modules/helm` path no longer exists)
- `terraform/terraform.md` points at the two new modules instead of the removed `helm` module


node/landing-page/src/App.tsx
```diff
@@ -78,11 +85,11 @@ function App(): JSX.Element {
       }
       if (config.useKeycloak) {
         keycloakConfig = {
-          url: config.keycloakAuthUrl,
+          url: config.keycloakAuthUrl!,
           realm: config.keycloakRealm!,
           clientId: config.keycloakClientId!
         };
-        const keycloak = Keycloak(keycloakConfig);
+        const keycloak = new Keycloak(keycloakConfig!);
```
jfaltermeier, line 92: I think the testing page needs to be adjusted as well. It does not build for me.

terraform/modules/theia-cloud/README.md
```diff
@@ -0,0 +1,46 @@
+# Theia Cloud Installation Module
+
+This module installs Theia Cloud components in a Kubernetes cluster via Helm.
+
+## Prerequisites
```
jfaltermeier, line 5: We should adjust the prerequisites a bit.
* HAProxy or NGINX Ingress Controller required
* Keycloak optional

terraform/modules/cluster-prerequisites/README.md
```diff
@@ -0,0 +1,291 @@
...
+- Cert-manager integration for automatic certificate generation
+- Optional self-signed ClusterIssuer for local development
+- Support for Minikube and generic Kubernetes clusters
+- Configurable HTTP relative path (e.g., `/keycloak/`)
+
+## Prerequisites
+
+The following components must be installed in your Kubernetes cluster before using this module:
+
+1. **nginx-ingress-controller** (if using ingress): For routing traffic to Keycloak
```
jfaltermeier: Is `nginx-ingress-controller` really a prerequisite for the whole module?

terraform/modules/cluster-prerequisites/main.tf
```diff
@@ -0,0 +1,562 @@
...
+  provisioner "local-exec" {
+    command = <<-EOT
+      echo "Waiting for Keycloak resource to report ready..."
+      kubectl wait keycloak/keycloak -n ${local.keycloak_ns} --for=condition=Ready --timeout=3m
+      echo "Waiting for Keycloak pods to be ready..."
+      kubectl wait pods -n ${local.keycloak_ns} -l app=keycloak --for=condition=Ready --timeout=3m
+      echo "Waiting for Keycloak service endpoint..."
+      kubectl wait --for=jsonpath='{.subsets[0].addresses[0].ip}' endpoints/keycloak-service -n ${local.keycloak_ns} --timeout=2m
+      echo "Keycloak is ready!"
+      echo "Waiting additional 5 for Keycloak authentication to be fully initialized..."
```
jfaltermeier:
```suggestion
      echo "Waiting additional 5s for Keycloak authentication to be fully initialized..."
```

jfaltermeier (changes requested):
Thanks, this looks very good already!

I tested a few combinations:

* Keycloak 21.1.2 + Theia Cloud from this branch -> works
* Keycloak 26 + Theia Cloud from this branch -> works
* Keycloak 26 + Theia Cloud 1.2.0 (the getting-started guides from this branch, unchanged) -> login fails

The last one is the only broken combination, and it only affects the getting-started guides. Existing deployments and the new landing page are fine.

I think it is the keycloak-js 20 landing page from the 1.2.0 chart running against Keycloak >= 22 which adds `iss` parts to the URL. Hardcoding the landing page to a KC26-compatible image (I used `theiacloud/theia-cloud-landing-page:1.3.0-keycloak26-next.92fcb7c2` for testing) fixed it.

I think we should update the getting-started guides using Theia Cloud 1.2.0 to use a KC26-compatible image like `theiacloud/theia-cloud-landing-page:1.3.0-keycloak26-next.92fcb7c2`. I think we can use this very image, maybe with a different tag like `1.2-KC26-compat`.

I also left a few inline comments, nothing major, except the testing page, which does not build with keycloak-js 26 atm.

jfaltermeier:
IP Review for Keycloak Update: https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/work_items/29451

terraform/modules/theia-cloud/README.md
lucas-koehler: Keycloak is not optional when using the module. I'll add a note that the module can be used when only installing crds and base though.

terraform/modules/cluster-prerequisites/README.md
lucas-koehler: No that was an oversight :see_no_evil: Good catch! In fact, the module can install the ingress itself :D

lucas-koehler:
@jfaltermeier Thanks for the review and tests :relaxed: 
I addressed the comments and pushed the KC26 compatibility landing page image and apply it in the theia-cloud module.

jfaltermeier (approved):
Thanks for addressing the remaining review comments. The testing page now builds with Keycloak 26, the landing page compatibility issue is resolved, and the Terraform documentation has been clarified.

I see no blockers. There could still be smaller issues in the Terraform configuration, but we can address those as we find them.

Approved.

PR #496, opened 2026-08-10
Upgrade to node 24 and typescript 5.9.3

### LINKED ISSUE #495: Update to Node 22 or later
Theia Cloud currently still uses Node 20 which is end of life and not compatible to latest npm versions.
We should upgrade to at least Node 22, preferably Node 24.

lucas-koehler:
- Upgrade node usage and @types/node to node 24
- Update Dockerfiles to use node 24
- Read node version from new .nvmrc files in CI workflows
- Upgrade Typescript to 5.9.3

E2E test run: https://github.com/eclipse-theia/theia-cloud/actions/runs/31398675288


.prompts/project-info.prompttemplate
jfaltermeier: This still mentions the old TypeScript version.

jfaltermeier (approved):
Thanks, lgtm overall. I just found one outdated doc where I added a comment.

.prompts/project-info.prompttemplate
lucas-koehler: Good catch! Fixed it

PR #497, opened 2026-08-10
ci: upgrade versions of all used GH actions

lucas-koehler:
E2E test run: https://github.com/eclipse-theia/theia-cloud/actions/runs/31398716528

### DIFF
--- a/.github/workflows/reusable-theia-extension.yml
+++ b/.github/workflows/reusable-theia-extension.yml
@@ -41,7 +41,12 @@ jobs:
           node-version-file: theia/.nvmrc
           registry-url: "https://registry.npmjs.org"
 
+      # setup-node's registry-url writes an .npmrc containing ${NODE_AUTH_TOKEN}. Since setup-node v7
+      # no dummy token is exported anymore and Yarn Classic fails on the unresolved placeholder.
+      # Publishing uses npm Trusted Publishing (OIDC) and does not need a real token here.
       - name: Install dependencies
+        env:
+          NODE_AUTH_TOKEN: dummy
         run: |
           sudo apt-get install libsecret-1-dev
           yarn --frozen-lockfile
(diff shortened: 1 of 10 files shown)


.github/workflows/reusable-theia-extension.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           fi
 
       - name: Set up Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v7
```
jfaltermeier, line 39: There is a breaking change in https://github.com/actions/setup-node#whats-new-in-v7 that might affect yarn 1. Not sure if we'll run into this

jfaltermeier: I'll create a temporary PR testing this

jfaltermeier: Yes, unfortunately there is a problem during `yarn install`: [https://github.com/eclipse-theia/theia-cloud/actions/runs/31792169620/job/94741306980?pr=498](https://github.com/eclipse-theia/theia-cloud/actions/runs/31792169620/job/94741306980?pr=498)
Minimal reproducer workflow;
https://github.com/eclipse-theia/theia-cloud/actions/runs/31792169620/workflow?pr=498

jfaltermeier (changes requested):
Thanks, the changes should mostly work as far as I can see. There is one breaking change that affects Yarn 1, though.

.github/workflows/reusable-theia-extension.yml
lucas-koehler: Good catch, thanks! It works when providing a dummy token during the install step. see run https://github.com/eclipse-theia/theia-cloud/actions/runs/31810329825/job/94799136695?pr=498 of the tmp pr.

lucas-koehler:
@jfaltermeier Thanks for the review and the great catch! I provided a fix.

jfaltermeier (approved):

PR #499, opened 2026-08-14
Add OpenShift session routing support and tests

jfaltermeier:
Related to https://github.com/eclipse-theia/theia-cloud-helm/pull/106

See `terraform/test-configurations/openshift.md` for testing locally
Especially check `Appendix A: Building and Pushing Custom Images` to build and get the local images into openshift

* Add OpenShift as a supported cloud provider, introduce OpenShift Route based session routing alongside the existing Ingress routing
* Refactor session routing behind a `SessionRoutingStrategy` abstraction, add OpenShift Route handling for lazy and eager sessions, preserve full session URLs and URL schemes, update proxy and client handling accordingly
* Add OpenShift focused Terraform test configurations, Keycloak setup, operator packaging fixes, and a MicroShift based e2e workflow covering the OpenShift deployment path

an older e2e test run is here: https://github.com/eclipse-theia/theia-cloud/actions/runs/25375110058


.github/workflows/e2e-tests-openshift.yml
```diff
@@ -0,0 +1,652 @@
...
+          exit 1
+
+      # ----------------------------------------------------------------
+      # Playwright tests
+      # ----------------------------------------------------------------
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
lucas-koehler, line 525: Should be upgraded to use node 24 like on main

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/routing/OpenShiftRouteRoutingStrategy.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    @Override
+    public synchronized String addSessionRouting(Session session, AppDefinition appDefinition, Service service,
+            int instance, String correlationId) {
+        ensureInitialized();
+        String routeName = computeInstanceRouteName(appDefinition, instance);
+        String hostname = computeInstanceHostname(appDefinition, instance);
+        return createRoute(routeName, hostname, session, appDefinition, service, correlationId);
+    }
```
lucas-koehler, line 144: ingressHostnamePrefixes is silently ignored here for OpenShift .
  IngressRoutingStrategy.java:171-176 adds one rule per prefix; the Route strategy creates exactly one Route for the base host and never reads the field. An AppDefinition using prefixes for webview wildcard hosts gets 404s on those hostnames with no log line. The class javadoc documents only the usePaths limitation.

Is this on purpose? If not, this can be a follow up IMO

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/routing/OpenShiftRouteRoutingStrategy.java
```diff
@@ -0,0 +1,290 @@
...
+
+    /** Package-private constructor for unit tests. */
+    OpenShiftRouteRoutingStrategy(TheiaCloudOperatorArguments arguments) {
+        this.arguments = arguments;
+    }
+
+    OpenShiftRouteRoutingStrategy() {
+    }
+
+    private synchronized void ensureInitialized() {
```
lucas-koehler, line 85: Route TLS/annotation config is cached for the operator's lifetime:
ensureInitialized() reads openshift-route-config once and never refreshes, and the chart's operator Deployment has no checksum annotation on that ConfigMap. helm upgrade --set ingress.tls=false leaves the operator emitting https:// session URLs until someone restarts it. Separately, cm.getData() at line 100 will NPE on a ConfigMap with no data block.

Not sure if we need to fix this but it could prevent unexpected bugs when running into this. Could also be a follow up if we want to do it in general but now now

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/routing/OpenShiftRouteRoutingStrategy.java
```diff
@@ -0,0 +1,290 @@
...
+            Service service, String correlationId) {
+        String namespace = client.namespace();
+        String serviceName = service.getMetadata().getName();
+
+        Route existingRoute = osClient.routes().inNamespace(namespace).withName(routeName).get();
+        if (existingRoute != null) {
+            LOGGER.info(formatLogMessage(correlationId, "Route '" + routeName + "' already exists with host '"
+                    + existingRoute.getSpec().getHost() + "'"));
+            return protocol() + existingRoute.getSpec().getHost() + "/";
+        }
```
lucas-koehler, line 184: Eager start reuses an existing Route without re-owning it:
The early return hands back the existing Route's URL while leaving ownerReferences pointed at the previous session. If that Session object is already gone (operator restart during cleanup), GC deletes the Route out from under the new session. Updating the owner reference on the reuse path would close it.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/util/TheiaCloudDeploymentUtil.java
```diff
@@ -40,18 +39,20 @@ public final class TheiaCloudDeploymentUtil {
...
-        return getSessionURL(host, ingressPathProvider.getPath(appDefinition, session));
+    /**
+     * Extract the host portion from a full URL by stripping the protocol scheme and
+     * trailing slash.
+     *
+     * @param url a URL such as {@code https://host/path/}
+     * @return the host+path without scheme or trailing slash, e.g.
+     *         {@code host/path}
+     */
+    public static String extractHost(String url) {
```
lucas-koehler, line 50: According to claude this is dead code:
Nothing in production calls it. The only references anywhere in the repo are the 5 tests in TheiaCloudDeploymentUtilTests.java:28-55, yet the CHANGELOG advertises it as a replacement utility. Drop it or use it.

Might be worth a check

terraform/values/valuesOpenShiftMonitor.yaml
```diff
@@ -0,0 +1,61 @@
+imagePullPolicy: IfNotPresent
+
+# Legacy app configuration (deprecated - use service.authToken instead)
+app:
+  id: asdfghjkl
+  name: Theia Blueprint
+
+demoApplication:
+  name: theiacloud/theia-cloud-activity-demo:1.2.0
```
lucas-koehler, line 9: Should probably point to the next version like the other configs. or is there a reason not to?
```suggestion
  name: theiacloud/theia-cloud-activity-demo:1.3.0-next
```

terraform/values/valuesOpenShiftMonitor.yaml
```diff
@@ -0,0 +1,61 @@
...
+hosts:
+  usePaths: false
+  configuration:
+    baseHost: apps-crc.testing
+    service: service
+    landing: try
+    instance: ws
+
+landingPage:
+  image: theiacloud/theia-cloud-landing-page:1.2.0
```
lucas-koehler, line 29: Should probably point to the next version like the other configs. or is there a reason not to?
```suggestion
  image: theiacloud/theia-cloud-landing-page:1.3.0-next
```

.gitignore
```diff
@@ -3,4 +3,8 @@
 openapitools.json
 .DS_Store
 node/e2e-tests/test-results/.last-run.json
-.prompts/task-contexts
\ No newline at end of file
+.prompts/task-contexts
+**/terraform.tfstate
+**/terraform.tfstate.backup
+**/.terraform/
+**/.terraform.lock.hcl
```
lucas-koehler, line 10: I think we should not ignore the lock files. or is there a special reason for this?

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/EagerSessionHandler.java
```diff
@@ -248,14 +238,19 @@ public boolean sessionAdded(Session session, String correlationId) {
             }
         }
 
-        /* adjust the ingress */
+        /* adjust the routing */
         String host;
         try {
-            host = updateIngress(ingress, serviceToUse, appDefinitionID, instance, port, appDefinition.get(),
+            host = routingStrategy.addSessionRouting(session, appDefinition.get(), serviceToUse.get(), instance,
```
lucas-koehler, line 244: Should `host` be renamed to `sessionUrl` to a void confusion? It no longer carries only the host.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/LazySessionHandler.java
```diff
@@ -268,16 +258,25 @@ protected boolean doSessionAdded(Session session, String correlationId) {
         createAndApplyDeployment(correlationId, sessionResourceName, sessionResourceUID, session, appDefinition,
                 storageName, arguments.isUseKeycloak(), labelsToAdd);
 
-        /* adjust the ingress */
+        /* adjust the routing */
         String host;
         try {
-            host = updateIngress(ingress, serviceToUse, session, appDefinition, correlationId);
+            host = routingStrategy.addSessionRouting(session, appDefinition, serviceToUse.get(), correlationId);
```
lucas-koehler, line 264: Should `host` be renamed to `sessionUrl` to a void confusion? It no longer carries only the host.

lucas-koehler (changes requested):
Hi @jfaltermeier , thank you for this extensive feature!
I tested the openshift support locally and it worked well for me.
I did not test the existing cloud providers but I think the e2e tests should cover that they still work :D
I have some comments in the code. Please have a look :)

.github/workflows/e2e-tests-openshift.yml
jfaltermeier: Fixed in follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577. The OpenShift E2E workflow now uses Node 24 through the same `.nvmrc` configuration as the current `main` CI workflow.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/routing/OpenShiftRouteRoutingStrategy.java
jfaltermeier: You're right that `ingressHostnamePrefixes` is currently not supported. The main use case is Theia's default `{{uuid}}.webview.{{hostname}}` endpoint. Supporting that natively requires an additional wildcard Route per session, wildcard Route admission enabled on the OpenShift Ingress Controller, and matching DNS and TLS certificate coverage.

We are not adding that feature in this PR. Instead, follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577 documents the current limitation and the `THEIA_WEBVIEW_EXTERNAL_ENDPOINT={{hostname}}` workaround, including the reduced origin isolation/security implications.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/routing/OpenShiftRouteRoutingStrategy.java
jfaltermeier: Fixed in follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577. The OpenShift client is still initialized lazily, but Route TLS and annotation settings are refreshed from `openshift-route-config` for routing operations instead of being cached for the operator lifetime. Missing ConfigMaps, missing `data`, and missing keys now fall back safely to TLS disabled and no annotations; malformed annotations retain warning/fallback behavior.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/routing/OpenShiftRouteRoutingStrategy.java
jfaltermeier: Fixed in follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577. Reused Routes now replace the previous Session owner reference with the current Session before the URL is returned, including the defensive cases where the existing Route has no or an empty owner-reference list. The returned protocol is also derived from the existing Route’s actual TLS configuration.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/util/TheiaCloudDeploymentUtil.java
jfaltermeier: Fixed in follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577. `extractHost()` was leftover from an earlier routing approach and had no production caller, so the method, its dedicated tests, and the changelog claim were removed.

terraform/values/valuesOpenShiftMonitor.yaml
jfaltermeier: Fixed in follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577. The OpenShift monitor values now use `1.3.0-next`, matching the existing monitor configuration.

terraform/values/valuesOpenShiftMonitor.yaml
jfaltermeier: Fixed in follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577. The landing-page image is now aligned to `1.3.0-next` as well.

.gitignore
jfaltermeier: Fixed in follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577. `.terraform.lock.hcl` is no longer globally ignored, while Terraform working directories and state files remain ignored. The OpenShift configuration lock files are committed.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/EagerSessionHandler.java
jfaltermeier: Fixed in follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577. The method-local variable was renamed to `sessionUrl` because it contains the complete URL. This is an internal rename and not a breaking API change.

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/handler/session/LazySessionHandler.java
jfaltermeier: Fixed in follow-up commit 1e4392565c7198e9dbfa1ad4d8f7ab358bb0b577. The corresponding local variable in the lazy handler was also renamed to `sessionUrl`; this is not a breaking change.

jfaltermeier:
Rebased and addressed the comments. Please have a look again

Green minikube run: https://github.com/eclipse-theia/theia-cloud/actions/runs/33387944232
Green openshift run: https://github.com/eclipse-theia/theia-cloud/actions/runs/33394425359?pr=499

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/routing/OpenShiftRouteRoutingStrategy.java
```diff
@@ -244,6 +284,34 @@ ResourceEdit.<Route> updateOwnerReference(0, Session.API, Session.KIND, session.
...
+            }
+        }
+        if (sessionOwner == null) {
+            sessionOwner = new OwnerReference();
+            ownerReferences.add(sessionOwner);
+        }
+        sessionOwner.setApiVersion(Session.API);
+        sessionOwner.setKind(Session.KIND);
+        sessionOwner.setName(session.getMetadata().getName());
+        sessionOwner.setUid(session.getMetadata().getUid());
```
lucas-koehler, line 312: Does this actually edit the the session owner on Kubernetes?
As I remember, we usually called an explicit edit via the kubernetes client.

lucas-koehler (changes requested):
Hi @jfaltermeier , thanks for the updates! The changes look pretty good to me now. I only have one concern inline. Please have a look :)

java/operator/org.eclipse.theia.cloud.operator/src/main/java/org/eclipse/theia/cloud/operator/routing/OpenShiftRouteRoutingStrategy.java
jfaltermeier: I think it should work, the helper itself only mutates the Route in memory, but it is called inside osClient.routes()...edit(route -> ...) in createRoute, and on the create path the mutated Route is passed to create().

Still, I pushed a commit that moves the helper to TheiaCloudHandlerUtil next to the other owner-reference helpers, adds javadocs, and wraps the edit in the same try/catch we use in EagerSessionHandler.

new e2e run: https://github.com/eclipse-theia/theia-cloud/actions/runs/35599078147

lucas-koehler: I see, that makes sense :+1:  Thanks for the extraction to the util and adding comments. That makes it clearer for the future :)

lucas-koehler (approved):
LGTM now, thanks 🚀

PR #501, opened 2026-08-25
fix(monitor): compare session secret in constant time

ofri-peretz:
## The bug

`isAuthorized()` gates five HTTP endpoints on a plain `===`:

```ts
const bearerToken = splitBearer[1];
if (bearerToken === sessionSecret) {
  return true;
}
```

- `node/monitor/src/util/util.ts` — used by `/activity`, `/popup`, `/message`
- `theia/extensions/monitor-theia/src/common/util.ts` — used by the same two backend modules

This is the CWE-208 pattern: comparison time depends on the contents of the
secret. I want to be precise about how much it actually leaks under V8, because
the honest answer is "less than the usual writeup claims, but not nothing."

Measured on Node 24, 21 rounds, median ns per comparison, probe differing from
the secret at a known offset:

```
secret 32B:  0 correct -> 13.8ns    8 -> 78.4ns   16 -> 73.2ns   31 -> 76.7ns
secret 64B:  0 correct -> 20.9ns   16 -> 174.7ns  32 -> 172.8ns  63 -> 118.0ns
```

So there is a sharp, trivially-measurable 5-8x step between "the leading bytes
are wrong" and "the leading bytes are right" — but **no byte-by-byte gradient
beyond that**, because V8 compares word-wise rather than one character at a
time. A byte-at-a-time recovery of a 32-byte random secret is therefore *not*
what this exposes; what it exposes is a reliable oracle for the leading word.

That is still worth closing. It is a real information leak on the check that
stands between the network and "stop this session" / "show this message
fullscreen", the correct primitive costs six lines, and the amount that leaks
is a V8 implementation detail rather than a guarantee — a future engine, a
shorter secret, or a secret with a predictable prefix all move it the wrong way.
I would rather not have the security of this check depend on how a particular
JIT happens to lay out `memcmp`.

## The fix

HMAC both sides under a per-process random key, then `crypto.timingSafeEqual`
the digests:

```ts
const COMPARE_KEY = randomBytes(32);

function secretEquals(actual: string, expected: string): boolean {
  const left = createHmac('sha256', COMPARE_KEY).update(actual).digest();
  const right = createHmac('sha256', COMPARE_KEY).update(expected).digest();
  return timingSafeEqual(left, right);
}
```

Hashing first is what makes this safe to drop in: `timingSafeEqual` throws when
the two buffers differ in length, and a bare length check would leak the
secret's length. Two 32-byte digests avoid both.

`src/common/util.ts` now imports `node:crypto`. It is only imported from
`src/node/modules/*`, and the extension's frontend entry is
`lib/browser/monitor-frontend-module`, so it does not reach a browser bundle.

## No behaviour change

The explicit `sessionSecret !== undefined` preserves the old
`string === undefined` semantics, so an unset `THEIACLOUD_SESSION_SECRET` still
fails closed exactly as before.

I ran the old and new implementations side by side over every combination of 18
authorization-header shapes (missing, empty, `Bearer` with no token, extra
spaces, wrong scheme, off-by-one tokens, the literal strings `undefined` /
`null` / `0`) and 9 secret values (including `undefined`, empty and
whitespace):

```
162 input pairs compared, 0 divergences
```

## The guard

A fix without a check is a fix that comes back. `eslint-plugin-node-security` is
registered in `node/configs` and `theia/configs` with **one rule** enabled —
`node-security/no-timing-unsafe-compare` — not the plugin's preset.

To be straight with you about how it decides: it matches the identifier being
compared against a list of secret-ish name patterns (`token`, `secret`, `key`,
`bearer`, …). In this file it is `bearerHeader` / `splitBearer` that trip it.
That is a heuristic, not a proof, and it means the rule can miss a
badly-named secret and can in principle flag a harmlessly-named non-secret.
The list is the `secretPatterns` option, so you can narrow or replace it:

```json
"node-security/no-timing-unsafe-compare": ["error", { "secretPatterns": ["secret", "bearer"] }]
```

Reverting either `util.ts` turns lint red:

```
node/monitor/src/util/util.ts
  10:11  error  CWE-208 OWASP:A02-Security CVSS:5.9 | Using === to compare secrets
         enables timing attacks. ...  node-security/no-timing-unsafe-compare
```

Pinned exactly (`5.2.2`). MIT, one transitive dependency (`@interlace/eslint-devkit`,
also MIT) — should pass the dash-licenses check.

## Verified

- `cd node/monitor && npm run lint` — clean; red on `git checkout upstream/main -- src/util/util.ts`
- `cd theia && npx lerna run lint` — no new findings (`config-store-example`'s
  `import/no-unresolved` is pre-existing and resolves once `yarn build` has run)
- `cd node/monitor && npm run webpack` — compiles; 2 warnings, the same 2 as on `main`
- `cd theia/extensions/monitor-theia && npm run build` — clean
- `tsc --noEmit` clean in both packages
- Neither package declares a `test` script, so there is no suite to regress
- `theia/yarn.lock` regenerated with yarn 1.22 (+12 lines), so `yarn --frozen-lockfile` still passes

## One thing you may want to know

Only `theia/` is linted in CI (`ci-theia.yml`). Nothing runs `npm run lint` under
`node/`, so the `node/monitor` copy of this bug would not have been caught there
even with the rule enabled. I have left the `node/` config wired up anyway since
`npm run lint` is the documented local command, but adding a lint step to a
`node/` workflow would close the gap.

Happy to split the lint-config part into its own PR, or drop it entirely and
leave just the two `util.ts` fixes, if you would rather not take the dependency.

### DIFF
--- a/node/configs/base.eslintrc.json
+++ b/node/configs/base.eslintrc.json
@@ -12,6 +12,7 @@
     "header",
     "import",
     "no-null",
+    "node-security",
     "react",
     "simple-import-sort",
     "prettier"
--- a/node/configs/errors.eslintrc.json
+++ b/node/configs/errors.eslintrc.json
@@ -12,6 +12,7 @@
     ],
     "guard-for-in": "error",
     "no-caller": "error",
+    "node-security/no-timing-unsafe-compare": "error",
     "no-eval": "error",
     "no-restricted-imports": [
       "error",
--- a/node/monitor/src/util/util.ts
+++ b/node/monitor/src/util/util.ts
@@ -1,4 +1,20 @@
+import { createHmac, randomBytes, timingSafeEqual } from 'node:crypto';
+
 import { getFromEnv, THEIACLOUD_SESSION_SECRET } from '../env-variables';
+
+/**
+ * Per-process key used only to length-normalise the two values before they are
+ * compared. Hashing both sides keeps the compared buffers the same size, so
+ * timingSafeEqual never throws on a length mismatch and the secret's length is
+ * not observable from response timing either.
+ */
+const COMPARE_KEY = randomBytes(32);
+
+function secretEquals(actual: string, expected: string): boolean {
+  const left = createHmac('sha256', COMPARE_KEY).update(actual).digest();
+  const right = createHmac('sha256', COMPARE_KEY).update(expected).digest();
+  return timingSafeEqual(left, right);
+}
 
 export function isAuthorized(req: any): boolean {
   const bearerHeader = req.headers['authorization'];
@@ -7,7 +23,7 @@ export function isAuthorized(req: any): boolean {
     if (splitBearer.length === 2) {
       const bearerToken = splitBearer[1];
       const sessionSecret = getFromEnv(THEIACLOUD_SESSION_SECRET);
-      if (bearerToken === sessionSecret) {
+      if (sessionSecret !== undefined && secretEquals(bearerToken, sessionSecret)) {
         return true;
       } else {
         console.debug('unauthorized');
--- a/node/package-lock.json
+++ b/node/package-lock.json
@@ -23,7 +23,7 @@
         "eslint-plugin-header": "^3.1.1",
         "eslint-plugin-import": "^2.26.0",
         "eslint-plugin-no-null": "^1.0.2",
-        "eslint-plugin-node-security": "5.2.2",
+        "eslint-plugin-node-security": "5.2.3",
         "eslint-plugin-prettier": "^4.0.0",
         "eslint-plugin-react": "^7.29.4",
         "eslint-plugin-simple-import-sort": "^7.0.0",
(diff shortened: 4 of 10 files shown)


jfaltermeier (commented):
Thanks for the fix. The timing-safe comparison itself looks reasonable, but we don't want to introduce `eslint-plugin-node-security` and its transitive dependency at this point, as these packages are not yet widely established.

I understand that you are the primary maintainer of these packages, but could you remove them from this contribution? We can then continue reviewing the timing-safe comparison itself.

One thing about contributing to Eclipse projects is that you need to sign the [Eclipse Contributor Agreement ECA](https://www.eclipse.org/legal/eca). This is a one-time requirement for contributing to Eclipse Foundation projects. Once signed, you can contribute to any Eclipse project. This is a bit inconvenient, I know, but we cannot change that.

To sign the ECA:

* Create an Eclipse Foundation account if you do not have one: [https://accounts.eclipse.org/user/register](https://accounts.eclipse.org/user/register)
* Then sign the ECA: [https://accounts.eclipse.org/user/eca](https://accounts.eclipse.org/user/eca)
* Make sure your Git commits use the same email address as your Eclipse Foundation account

